Issue #1, opened 2015-12-20
Crash when de_DE locale is not installed


alexis-via:
I am trying to use invoice2data on my Ubuntu 15.10:

```
% python -m invoice2data.main invoice2data/test/pdfs --debug
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/alexis/new_boite/dev/invoice2data/invoice2data/main.py", line 67, in <module>
    res = extract_data(join(args.file_folder, f))
  File "/home/alexis/new_boite/dev/invoice2data/invoice2data/main.py", line 34, in extract_data
    output[k] = str2date(raw_date)
  File "invoice2data/date_parser.py", line 33, in str2date
    with setlocale(l):
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "invoice2data/date_parser.py", line 18, in setlocale
    yield locale.setlocale(locale.LC_ALL, name)
  File "/usr/lib/python2.7/locale.py", line 579, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting
zsh: exit 1     python -m invoice2data.main invoice2data/test/pdfs --debug
```
As I am a French user, my default locale is French:
```
% locale
LANG=fr_FR.UTF-8
LANGUAGE=fr_FR
LC_CTYPE="fr_FR.UTF-8"
LC_NUMERIC="fr_FR.UTF-8"
LC_TIME="fr_FR.UTF-8"
LC_COLLATE="fr_FR.UTF-8"
LC_MONETARY="fr_FR.UTF-8"
LC_MESSAGES="fr_FR.UTF-8"
LC_PAPER="fr_FR.UTF-8"
LC_NAME="fr_FR.UTF-8"
LC_ADDRESS="fr_FR.UTF-8"
LC_TELEPHONE="fr_FR.UTF-8"
LC_MEASUREMENT="fr_FR.UTF-8"
LC_IDENTIFICATION="fr_FR.UTF-8"
LC_ALL=
```

If I run "locale-gen", I see that Ubuntu generate all fr_* and en_* locales.

[...]

PR #3, opened 2015-12-21
Prepare lib for Odoo PDF import

alexis-via:
This PR contains my work of these last days on this lib to make it work with my module on Odoo PDF import.

The fact is that I modified several different things:
- try to fix bug #1 
- add UTF-8 headers
- update date parsing
- add templates for several French companies

This is still a bit a "work in progress", as I'll continue to work on this branch in the coming days. But this PR should help you see what changes I'm doing and maybe get your feedback if you disagree on some changes I make.


manuelRiel:
Awesome. Can you also add some sample files, so I can try out the new templates? And maybe add them to a test folder? Just drop them here. No need to add them in the PR.

alexis-via:
@manuelRiel 
I'm glad that you like my work.
FYI, the code of my Odoo module that use your invoice2data lib is here: https://github.com/akretion/account-invoicing/tree/8.0-add-pdf-invoice-import/account_invoice_pdf_import (AGPL licence)

About the template files: I should be able to add a few, but I got some of them via friends and I told them that I would keep their invoice confidential. Some of them are phone bill invoices with detailed call records, so you can understand that it's not a good idea for privacy reasons to add it to the "test" folder... but in the meantime I understand the need to have a test suite to check the templates.

alexis-via:
@manuelRiel FYI, I just announced the availability of the Odoo module "account_invoice_pdf_import" on my company blog: http://www.akretion.com/blog/akretions-christmas-present-for-the-odoo-community So you will know the story behind our work !

I hope it will bring more contributors to your great project.

manuelRiel:
Thanks for the contribution. I added a license, merged your code and added you to contributors. And will upload to Pypi.

I'll probably work on this again when I do my own accounting. What other features would be interesting? The main goal is to recognize new invoices without templates. But to train a ML algo, I don't have enough sample data right now.

Issue #9, opened 2016-01-27
Cannot use extract_data() directly


alexis-via:
OK, I found the time to investigate the issue. After the commits of this week-end, it still works well when you use invoice2data from the command line, but it doesn't work any more when you use it as a python lib.

 In the README file, it says:

```
from invoice2data import extract_data

result = extract_data('path/to/my/file.pdf')
```

This doesn't work any more, because extract_data has one additionnal arg. This additional arg must be the result of a call to the read_templates() method, which is done in the main() method before calling extract_data().

I don't know what you have in mind about this. I see two options:
1) we update the README to explain that users have to call read_templates() first and then call extract_data()... which is a bit more work for our lib users than before. 
2) we adapt the code to call read_template inside extract_data if the second arg is empty

manuelRiel:
option 2

alexis-via:
@manuelRiel I made a PR that implements option 2

PR #10, opened 2016-01-27
Fix bug #9 : allow to use extract_data() with a single arg

alexis-via:
This PR is just a proposal. It may not be the best idea and may not be what you had in mind when you made the changes in the code this WE.


manuelRiel:
Looks fine. I would have done it like this. Can merge after CircleCI is done.

alexis-via:
OK. CircleCI is green, so I merge.

PR #22, opened 2017-01-02
[ADD] proposal to support invoice lines

hbrunn:
this is a proposal to support lines in invoices. With the proposed changes to `de.qualityhosting.yml`, it parses the demo invoice okayish.

My questions are:

- are you interested in the first place to support invoice lines?
- what do you think about the ordered dict comment in the code?
- this is quite sensitive to left out values/multiline texts. Would you prefer fixed width parsing?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ With the flexible template system you can:
 Go from PDF files to this:
 
 ```
-{'date': (2014, 5, 7), 'invoice_number': '30064443', 'amount': 34.73, 'desc': 'Invoice 30064443 from QualityHosting'}
+{'date': (2014, 5, 7), 'invoice_number': '30064443', 'amount': 34.73, 'desc': 'Invoice 30064443 from QualityHosting', 'lines': [{'price': 42.0, 'desc': u'Small Business StandardExchange 2010\nGrundgeb\xfchr pro Einheit\nDienst: OUDJQ_office\n01.05.14-31.05.14\n', 'pos': u'7', 'qty': 1.0}]}
 {'date': (2014, 6, 4), 'invoice_number': 'EUVINS1-OF5-DE-120725895', 'amount': 35.24, 'desc': 'Invoice EUVINS1-OF5-DE-120725895 from Amazon EU'}
 {'date': (2014, 8, 3), 'invoice_number': '42183017', 'amount': 4.11, 'desc': 'Invoice 42183017 from Amazon Web Services'}
 {'date': (2015, 1, 28), 'invoice_number': '12429647', 'amount': 101.0, 'desc': 'Invoice 12429647 from Envato'}
--- a/README.md
+++ b/README.md
@@ -95,12 +95,17 @@ options:
   currency: HKD
   date_formats:
     - '%d/%m/%Y'
+lines:
+    start: Detail
+    end: \* May include estimated US sales tax
+    first_line: ^    (?P<description>\w+.*)\$(?P<price_unit>\d+\.\d+)
+    line: (.*)\$(\d+\.\d+)
+    last_line: VAT \*\*
 ```
 
 ## Roadmap and open tasks
 
 - tutorial and documentation for template options.
-- parse invoice items.
 - integrate with online OCR?
 - try to 'guess' parameters for new invoice formats.
 - can apply machine learning to guess new parameters?
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -36,6 +36,43 @@ All the regex `fields` you need extracted. Required fields are `amount`, `date`,
 
 You will need to understand regular expressions to find the right values. If you didn't need them in your life up until now (lucky you), you can learn about them [here](http://www.zytrax.com/tech/web/regex.htm) or [test them here](http://www.regexr.com/). We use [Python's regex engine](https://docs.python.org/2/library/re.html). It won't matter for the simple expressions we need, but sometimes there are subtle differences when e.g. coming from Perl.
 
+### Lines
+The `lines` key allows you to parse invoice items. Mandatory are regexes `start` and `end` to figure out where in the stream the item table is located. Then the regex `line` is applied, and supposed to contain named capture groups. The names of the capture groups will be the field names for the parsed item. If we have an invoice that looks like
+
+```
+some header text
+
+the address, etc.
+
+  Item        Discount      Price
+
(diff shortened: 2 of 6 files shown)


m3nu:
Thanks for the commit, @hbrunn.  To respond to your questions:

- Yes, it's clearly a nice addition to parse invoice items. I may re-arrange some code to keep it organized, but it's clearly useful.
- If there is [a way](http://stackoverflow.com/questions/5121931/in-python-how-can-you-load-yaml-mappings-as-ordereddicts) to read the YAML as `orderedDict`, we should do it. I don't see a downside to it.
- White space is hard to deal with and usually we remove it (`remove_whitespace: true`). If I were to implement this without looking at your way or testing it, I'd try to make a regex for lines/invoice items with different capture groups and match that to the whole file. E.g. `(number)\s{,4}(desc)\s{,4}(qtty)\s{,4}(total)`. Not sure that's always possible of course, but should make it more reliable and mentally easier to add.

hbrunn:
great to hear. The ordering made its own PR and I'll keep working on the line parsing. One big expression with groups was my first idea too, but when we have to deal with partial lines/missing values, this gets quite complicated. And most real live invoices I come across at least have potentially multiple lines per item.

In the end, this will probably come in two flavors:

```yml

lines:
    fields:
        field1: 'regex'
        field2: 'regex'

```

for simple cases, but also

```yml

lines:
    fields:
        field1:
            regex: 'regex'
            start_hint: 42
            end_hint: 50
        field2:
            regex: 'regex'
            start_hint: 42
            end_hint: 50
```

to allow giving rough positions in the table in order to handle multiline fields correctly. Fixed width only seems a bit too inflexible.

m3nu:
Thanks. Let's see how it goes. I'll test it on some invoices as well and see if it's feasible.

hbrunn:
@m3nu after a bit of experimentation, I think we're best off with multiple line regexes. I now allow `first_line`, `line`, `last_line` regexes, which are checked in this order. Then the first matches, it's a new line, when the last matches, that's the end of a line, and if `line` matches, its values are appended to the previous line. Up to now, I could use this to parse a couple of real world invoices, including the test one, which results in
```
{
    'amount_untaxed': 34.73, 
    'lines': [
        {
            'price': u'3,89', 
            'desc': u'Small Business StandardExchange 2010\nGrundgeb\xfchr pro Einheit\nDienst: OUDJQ_strukan\n01.05.14-31.05.14',
            'pos': u'1', 
            'qty': u'1'
        }, 
        {
            'price': u'5,39', 
            'desc': u'Small Business QualityExchange 2010\nGrundgeb\xfchr pro Einheit\nDienst: OUDJQ_schneider\n01.05.14-31.05.14',
            'pos': u'2', 
            'qty': u'1'
        }, 
        {
            'price': u'5,39', 
            'desc': u'Small Business QualityExchange 2010\nGrundgeb\xfchr pro Einheit\nDienst: OUDJQ_minar\n01.05.14-31.05.14',
            'pos': u'3', 
            'qty': u'1'
        }, 
        {
            'price': u'5,39', 
            'desc': u'Small Business QualityExchange 2010\nGrundgeb\xfchr pro Einheit\nDienst: OUDJQ_mayr\n01.05.14-31.05.14',
            'pos': u'4', 
            'qty': u'1'
        }, 
        {
            'price': u'5,39', 
            'desc': u'Small Business QualityExchange 2010\nGrundgeb\xfchr pro Einheit\nDienst: OUDJQ_jenewein\n01.05.14-31.05.14',
            'pos': u'5', 
            'qty': u'1'
        },
        {
            'price': u'5,39', 
            'desc': u'Small Business QualityExchange 2010\nGrundgeb\xfchr pro Einheit\nDienst: OUDJQ_jauernik\n01.05.14-31.05.14\nQualityHosting AG - Uferweg 40-42 - D-63571 Gelnhausen\niViveLabs Ltd.\n93B Sai Yu Chung\nYuen Long, N.T.\nHong Kong\nPos.            Menge      Beschreibung                                                            Rabatt %     VK-Preis     Zeilenbetrag\nOhne      Ohne MwSt.\nMwSt.',
            'pos': u'6', 
            'qty': u'1'
        }, 
        {
            'price': u'3,89', 
            'desc': u'Small Business StandardExchange 2010\nGrundgeb\xfchr pro Einheit\nDienst: OUDJQ_office\n01.05.14-31.05.14\n',
            'pos': u'7', 
            'qty': u'1'
        }
    ], 
    'currency': 'EUR', 
    'amount': 34.73, 
    'date': datetime.datetime(2014, 5, 7, 0, 0), 
    'invoice_number': u'30064443', 
    'vat': u'DE 232 446 240', 
    'desc': u'Invoice 30064443 from QualityHosting AG'
} 
```
So item pos 6 is a bit botched because the description also includes the footer, but this can be fixed with replacements on a per yml file level in my opinion.

If you agree, I'd update the readme and the tutorial to use this, then we can merge, right?

PS: Would you merge a PR that changes the coding style to adhere to pep8/flake8?

m3nu:
Looks good. `price`, `pos` and `qty` should be ints and floats, I guess.

Feel free to PR pep8. We're not hostile to it, just didn't apply it yet.

hbrunn:
hmm, with the types, wouldn't it be sensible to add this to the yml file? The way I'd like to use this library is with basically arbitrary field names, and the consuming application will take care of type conversions.

How about: If a capture group's name contains a colon, the right hand side is the type the value should be coerced to? So that `(?P<pos>\d+)` gives you `'42'`, `(?P<pos:int>\d+)` yields `42` and `(?P<pos:float>\d+)` `42.0`? (this looks so much like werkzeug it should feel familiar to python programmers)

I'll update code and readme with your reply tomorrow, and propose pep8 probably after merging and beefing up the tests a bit.

hbrunn:
now it seems fine to me. The werkzeug-style type specifiers don't work because a capture group's name must be a python identifier. Not much choice here.
So I added a list `types` under the `lines` key, where you can set the type to coerce to per field. If you set nothing, it will stay a string

hbrunn:
@m3nu added some text for the readme, now I consider this finished

m3nu:
Thanks a lot. Will check it out, make sure your credits are added and then merge or get back to you.

m3nu:
Merged. Looks fine and only runes when `lines` setting is defined. Now we just need some templates to use it.

PR #29, opened 2017-05-23
read templates in alphabetical order

brakhane:
Try templates in alphabetic order. Previously, the entries were parsed
in arbitrary order which could lead to problems when more than one
template matches.

Reading the templates in alphabetical order is useful for defining a
"fallback" template that tries to extract information from unknown
invoices, while the specific templates can be used where necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ See `invoice2data/templates` for existing templates. Just extend the list to add
 
 Templates are based on Yaml. They define one or more keywords to find the right template and regexp for fields to be extracted. They could also be a static value, like the full company name.
 
+Template files are tried in alphabetical order.
+
 We may extend them to feature options to be used during invoice processing.
 
 Example:
--- a/invoice2data/template.py
+++ b/invoice2data/template.py
@@ -1,5 +1,5 @@
 """
-This module abstracts templates for invoice providers. 
+This module abstracts templates for invoice providers.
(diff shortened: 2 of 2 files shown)


m3nu:
Is this really needed? Generally only ONE template should match. I usually use the VAT number which should be unique.

brakhane:
@m3nu The use case we have for this is the following: We get many invoices from many suppliers and noticed that most of those invoice templates tend to look similar, for example one supplier might have the configuration 
```
amount: Invoice Amount (\d+)

m3nu:
Wouldn't it be better to inherit from a base template to avoid repeating the same configs?

brakhane:
The main point is that non-technical people should be able to import a new, unknown invoice. If the fallback template works, there is nothing left to do, only if it doesnt do technical people have to get involved.

AIUI, inheriting from a base template would still require a template be created for each supplier.

m3nu:
OK. I see the benefit. So there would be a "ZZDefault" template that gets tried last and has no keywords at all? This would be great of course. Do you have that template somewhere?

brakhane:
Actually, it has a single keyword: "."; since this matches every character, it will always be tried.

The template is still WIP and very hacky ("works for us, most of the time"), and tuned for German invoices. Once it matures, I can create a PR to include it.

PR #36, opened 2017-06-15
Do not use pytesseract

maethor:
Hi,

It is "not reliable at the moment" because of PyTesseract, because it converts the tiff in bmp before passing it to tesseract, and a lot of quality is lost in the process. You don't need it. You can directly use the tesseract command in subprocess, which is simplifying the code a lot (you don't need a tempfile, you don't need Image, etc.).

It doesn't solve everything, because tesseract is not perfect, but I tested this patch on good scans and it is working a lot better than before.

### DIFF
--- a/invoice2data/in_tesseract.py
+++ b/invoice2data/in_tesseract.py
@@ -1,35 +1,17 @@
 # -*- coding: utf-8 -*-
-# from wand.image import Image as WandImage
-from PIL import Image
-import pytesseract
-import os
 import subprocess
-import tempfile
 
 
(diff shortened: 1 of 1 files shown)


m3nu:
Cool. In combination with removing white space this should give good results on scan.s

m3nu:
So if no text is found in the PDF, it will fall back to tesseract again.

I'll adapt that and add tesseract as optional install to the instructions.

maethor:
Thank you @m3nu!

PR #39, opened 2017-07-27
support empty matches in lines

brakhane:
Allow group matches to be optional, for example

  line: (?P<date>\d{2}\.\d{2})?\s(?P<name>.*)

the "date" group can be missing.


m3nu:
Sorry it took so long.

PR #41, opened 2017-08-10
Added template for Digi Key invoices in Germany

D4N:
I have created a template for invoices from Digi Key Germany.

It can extract most of the information from the lines and works with multi-page invoices. Unfortunately I have failed to successfully extract the product name (see comments in the template). Also this only works with credit card payments, due to the format of the invoice (and I don't have any invoices that weren't paid by credit card, so I can't test it).


m3nu:
Sorry for the delay.

PR #48, opened 2017-09-21
Add more extensibility for the conditions that make a field a date

gfcapalbo:
Currently a date field is identified if starts with "date" . I extend this a little bit with a lambda function because i also need requested_date in my  templates to be evaled like a date.
This looks like a nice way to extend that depending on one's needs for date fields

### DIFF
--- a/invoice2data/template.py
+++ b/invoice2data/template.py
@@ -27,9 +27,6 @@ OPTIONS_DEFAULT = {
     'line_separator': r'\n',
 }
 
-# add to this lambda if you want more fields parsed as date field
-is_date_field = lambda x: x.startswith('date') or x == 'requested_date'
-
 def read_templates(folder):
     """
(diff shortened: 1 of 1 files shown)


alexis-via:
Maybe we should just write:
`if 'date' in k:`

alexis-via:
or, if we don't want to make it too wide:
`if k.startswith('date') or k.endswith('date'):`

gfcapalbo:
I think start and end is enough. No lambda. pushing

hbrunn:
I still think there should be an extra key `types` for templates that allows arbitrary typecasts. This already [happens for lines](https://github.com/m3nu/invoice2data/blob/master/invoice2data/template.py#L262), and you can simply use [coerce_type](https://github.com/m3nu/invoice2data/blob/master/invoice2data/template.py#L132) then

PR #49, opened 2017-09-25
Skip empty sublines

gfcapalbo:
Found a template with empty sub-lines in the line itself. this fixes that.

e.g.

      "Line that corresponds to the template first_line"
            \s\s\s\s\\n
      "Line that corresponds to the template line"
          \s\\s\s\s\\n
      " another line that corresponds to the template line"

### DIFF
--- a/invoice2data/template.py
+++ b/invoice2data/template.py
@@ -213,7 +213,7 @@ class InvoiceTemplate(OrderedDict):
             self['lines']['first_line'] = self['lines']['line']
         for line in re.split(self.options['line_separator'], content):
             # if the line has empty lines in it , skip them
-            if  re.search(r'''^\s*\n$''', line) or not line:
+            if not line.strip('').strip('\n') or not line:
                 continue
             if 'first_line' in self['lines']:
                 match = re.search(self['lines']['first_line'], line)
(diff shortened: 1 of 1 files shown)


invoice2data/template.py
```diff
@@ -212,6 +212,9 @@ def extract_lines(self, content, output):
                 'last_line' not in self['lines']:
             self['lines']['first_line'] = self['lines']['line']
         for line in re.split(self.options['line_separator'], content):
+            # if the line has empty lines in it , skip them
+            if  re.search(r'''\s+\n''', line) or not line:
```
gfcapalbo: covers the case of an empty line or a line with only spaces

invoice2data/template.py
```diff
@@ -212,6 +212,9 @@ def extract_lines(self, content, output):
                 'last_line' not in self['lines']:
             self['lines']['first_line'] = self['lines']['line']
         for line in re.split(self.options['line_separator'], content):
+            # if the line has empty lines in it , skip them
+            if  re.search(r'''^\s*\n$''', line) or not line:
```
hbrunn: no need for the big guns: https://docs.python.org/2/library/string.html#string.strip

PR #60, opened 2018-02-10
replaced raw_date with res_find[0] in template.py file

rsenwar:
raw_date is not defined anywhere in the template.py file. Date parsing fails for res_find[0], Hence error logger must show res_find[0]. 
-Tested on My own Template.
@dpocock


m3nu:
Right. This error message would fail. Merging.

PR #65, opened 2018-02-15
Command line argument added for choosing output csv file

rsenwar:
@m3nu Need suggestions regarding the changes I made, you can edit/change accordingly.
I have added the command line argument for choosing output csv file, using -o or --output. If no arguments given, writes to 'invoices-output.csv' by default.


m3nu:
You need to rebase first. This PR has some old commits.

PR #68, opened 2018-02-15
Add optional csv naming feature

duskybomb:
using `--csv-name` or `-n` the user can change the default name.
Example:
`invoice2data -n invoice_csv inv.pdf` saves the csv as `invoice_csv.csv`
@dpocock

### DIFF
--- a/invoice2data/main.py
+++ b/invoice2data/main.py
@@ -62,6 +62,9 @@ def main():
     parser.add_argument('input_files', type=argparse.FileType('r'), nargs='+',
                         help='File or directory to analyze.')
 
+    parser.add_argument('--csv-name', '-n', dest='change_csv_name',
+                        help='Custom name for csv to be saved.')
+
     args = parser.parse_args()
 
     if args.debug:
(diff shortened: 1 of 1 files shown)


invoice2data/main.py
```diff
@@ -90,7 +93,15 @@ def main():
                     date=res['date'].strftime('%Y-%m-%d'),
                     desc=res['desc'])
                 shutil.copyfile(f.name, join(args.copy, filename))
-    invoices_to_csv(output, 'invoices-output.csv')
+    if args.change_csv_name:
```
m3nu: Instead of this long `if`-condition it may be possible to set a default value directly at `parser.add_argument`?

See here for many examples: https://docs.python.org/3/library/argparse.html

m3nu: @duskybomb ^^

duskybomb: Done

m3nu: Looks better. I still wanna make some changes. Will merge later.

Issue #70, opened 2018-02-16
Cannot convert to CSV


duskybomb:
```
File "/home/harshit/anaconda3/lib/python3.6/site-packages/invoice2data-0.2.73-py3.6.egg/invoice2data/main.py", line 97, in main
    invoices_to_csv(output, csv_output_name)
NameError: name 'csv_output_name' is not defined
```
Need to replace `csv_output_name` with `args.csv_output_name`

PR #71, opened 2018-02-16
Fix for Issue #70

duskybomb:
replaced `csv_output_name` with `args.csv_output_name`
Issue #70


m3nu:
Good find. Already an important lesson here: never add code without tests.

PR #73, opened 2018-02-17
some import paths, filenames updated and json output added

rsenwar:
some imports and file mismatches corrected.


rsenwar:
I have tested the project after the above changes, works fine.

PR #76, opened 2018-02-18
minor fixes

rsenwar:


invoice2data/main.py
```diff
@@ -7,15 +7,15 @@
 from os.path import join
 import logging
 
-from .input import pdftotext
-from .input import pdfminer
-from .input import tesseract
+from input import pdftotext
+from input import pdfminer
+from input import tesseract
```
duskybomb: Have you tested this commit @rsenwar ?
Doesn't seem to work in my environment.

m3nu: Agree. This won't work. I just undid those changes earlier.

rsenwar: I have tested it, works fine.
I get an error without these changes.

'from .input import pdftotext
ValueError: Attempted relative import in non-package'

rsenwar: Yeah, I understood, I have run that as a whole package.

m3nu: Are you running the main-file directly? That won't work and shouldn't work. Needs to run as module. See [here](https://stackoverflow.com/a/28154841) for a very exhaustive explanation.

tl;dr:
- They recommend absolut imports. To get those you still need to add the package name.
- Many people including me like relative imports, prefixed with a dot. The only downside is that you can't run a file directly. See *solution 1* in the link above.

m3nu: So the way to run it is either:

- `python -m invoice2data.main -h`
- `invoice2data` (linked cli tool)

This won't work:
- `python invoice2data/main.py`

rsenwar: My bad! Yeah It works fine.

m3nu: Great to hear. We're all here to learn. I also had to look up the precise reason. I guess some years ago I just got used to relative imports and ran every bigger project as module. Then forgot  the exact reason.

PR #81, opened 2018-02-20
This fixes ResourceWarning: unclosed file

student-06:
I am just opening files with context manager (```with``` syntax) in ```template.py```. Files will be closed automatically by context manager. Thanks. @dpocock


m3nu:
Thanks for your commit. This [change is already added](https://github.com/m3nu/invoice2data/blob/v2/invoice2data/extract/loader.py#L46) in our dev-branch, but there are still resource-warning.

Maybe you can find the cause for those in the [dev-branch](https://github.com/m3nu/invoice2data/tree/v2).

Issue #86, opened 2018-03-08
Lack of templates for Indian invoice issuers


AvatarSenju:
There is much lack of templates with regards to India companies which have high retail flux like flipkart and snapdeal and there may be differences in the Indian and other variants of the included firms like amazon India and Microsoft India.

PS:
So as a Solution and improvement to invoice2data I want to take this up as a project for GSoC 2018

m3nu:
You are correct. Most templates are EU and US-focused for now. Feel free to add new templates for India. Just make a new folder with the top-level domain.

AvatarSenju:
@m3nu I have submitted a draft proposal for this project for the Summer of Code with Google and I would humbly request you to review it.
It would be really helpful if you could provide some medium for personal contact or ping me on my email : anushrut5@gmail.com

rmilecki:
Flipkart template has been merged, if you want / have more templates, please open a pull request.

PR #90, opened 2018-03-12
template added : flipkart   #86

AvatarSenju:
First of the few templates being added.
Have checked against sample invoice and real invoice.
Any improvements are welcomed.


m3nu:
Could you add a sample PDF as well, so we can test it. Maybe even add it to the test cases.

Private information can be redacted in e.g. Adobe Acrobat or some similar PDF software.

AvatarSenju:
[OD304175096047380001.pdf](https://github.com/m3nu/invoice2data/files/1803329/OD304175096047380001.pdf)

sample pdf

AvatarSenju:
added test pdf for com.flipkart.WSRetail.yml

Issue #92, opened 2018-03-14
Better error messages when dependencies for Tesseract are missing


utsav666:
there is an issue with tesseract.py it gives error in line no 7 """"subprocess.Popen(convert.split(' '), stdout=subprocess.PIPE)"""" I repllace the line with this """"subprocess.Popen(convert.split(' '), stdout=subprocess.PIPE,shell=True)"""" but it gives a new error 
""""-density: 1: -density: convert: not found
Error in pixReadMem: size < 12
Error during processing."""""

m3nu:
This error just means that the `imagemagick` library is not installed. This prerequisite should be mentioned in the README. Not sure if it is.

The Tesseract module didn't get much love recently. I'm sure there are some things to improve. Like properly checking for installed prerequisites.

m3nu:
This is almost fixed. Commit still needs some work as the conditions are wrong and there is a typo in the import statement.

duskybomb:
Yes, I don't know how it happened. Make the correction though.

PR #93, opened 2018-03-14
Updated README.md

AvatarSenju:
With respect to issue #92 I have updated the [README.md](https://github.com/m3nu/invoice2data/blob/master/README.md) with a bit of information.
I would like to suggest a new section of prerequisites in [README.md](https://github.com/m3nu/invoice2data/blob/master/README.md) as @m3nu suggested.
If I could be provided with some of the unusual requirements then I can edit the [README.md](https://github.com/m3nu/invoice2data/blob/master/README.md) to make it even more readable and simple for new users and contibutors

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ If possible get the latest [xpdf/poppler-utils](https://poppler.freedesktop.org/
 
 2. Install `invoice2data` using pip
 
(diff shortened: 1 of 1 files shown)


m3nu:
- Typo in the sentence.
- Instructions seem to vague. Either something is needed or it's not needed.
- There should be code to check for missing dependencies and give a good error message, like we already have in `pdftotext` or the other default input library.

PR #94, opened 2018-03-15
Update README.md

duskybomb:
Added examples for `input-reader`, `output -format` and `output-name`


m3nu:
Thanks. Next challenge could be to get the OCR working better, now that we officially support it.

duskybomb:
@m3nu any suggestions on how I should approach this OCR challenge?

m3nu:
1. Get some test data. Maybe there is a public data set.
2. Preprocessing with imagemagick to make life easier for the OCR.
3. Tune OCR settings.
4. Test with regex rules.

Issue #97, opened 2018-03-18
Unicode Decode error while opening some templates


sanjay-94:
I am getting the following error while opening some templates. 

`templates += read_templates(os.path.abspath(args.template_folder))
  File "/Users/sanjay/code/invoice2data/invoice2data/invoice2data/extract/loader.py", line 47, in read_templates
    tpl = ordered_load(template_file.read())
  File "/Users/sanjay/PycharmProjects/test/venv/bin/../lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]`

`UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 40: ordinal not in range(128)`

I believe it is something to do with opening the templates in the below lines. 

`with open(os.path.join(path, name)) as template_file:`

When I change the line to 

`with open(os.path.join(path, name), encoding="utf-8") as template_file:`

the error **disappears**. Has the error got something to do with the IDE or the OS? And should we standardise the decoding codec through code, like enforcing it to utf-8?

m3nu:
We already have a similar issue #32.

If you like you can do a pull request for your suggested fix. Instead of passing the `encoding` parameter, you can also consider using the `codec` module, if it's supported in Python 2 and 3. It may be even more reliable than hard-coding a certain encoding.

sanjay-94:
I have made the changes using codec and chardet modules. Sent the PR.

sanjay-94:
Issue has been fixed.

PR #99, opened 2018-03-19
Fix for Issue #97 : Unicode Decode error while opening some templates

sanjay-94:
Used codecs and chardet module to dynamically determine the character of the template and open with appropriate decoding.


sanjay-94:
Can this be merged?

m3nu:
I'm not sure the `chardet` package is even needed. Let's see..

PR #104, opened 2018-03-31
allow type coercion for fields

rost314:
Not sure if this is considered useful since there is a possibility to force basic type coercion by prefixing the fields with 'date' or 'amount'. But since explicit is better than implicit, maybe it is.

### DIFF
--- a/invoice2data/extract/invoice_template.py
+++ b/invoice2data/extract/invoice_template.py
@@ -169,6 +169,10 @@ class InvoiceTemplate(OrderedDict):
             if plugin_keyword in self.keys():
                 plugin_func.extract(self, optimized_str, output)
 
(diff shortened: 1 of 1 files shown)


m3nu:
It should be automagical where possible. For edge cases, I see some use in specifying it. This would probably need more complicated field definitions. So each field to extract would have a name, regex, type, etc defined.

I also don't see an example template or any documentation for this feature. This would be useful too before merging it.

m3nu:
The feature in general is useful, but for a full PR, there should be

- updated documentation
- sample template
- new test template

The code is also a bit hard to understand/read.

If you are interested in this feature, I suggest you open an issue and we have a discussion there. Then do a more complete PR.

PR #112, opened 2018-04-29
Latest working tests

duskybomb:
Tested on python2 and python3, most probably this will work as expected.


m3nu:
ok. Will test this. Another option would be to parse the JSON before comparing. Less potential for false results.

m3nu:
Here my issues with this code:

- The json-file can't be version-dependent. That would be crazy. You need to parse it field-by-field and compare that instead of the file byte-by-byte.
- The main part of the test is commented out. So it doesn't do anything.
- You specify each PDF individually. This doesn't scale as people add new PDFs. You need a proper way to e.g. save the PDF and json with the same name, but different extension. Then compare all files that have comparison data available.
- This test concerns the whole library, not just the CLI. So it can go to a higher level.

PR #117, opened 2018-05-12
Better message when tesseract dependencies missing

duskybomb:
A quick fix for better message when tesseract dependencies are missing.
Solves #92


m3nu:
Can't use this for several reasons:

1. A simple `except` without specifying the precise exception will catch all errors and make debugging very hard. This needs to be more precise.
2. Testing for the presence of the binary and maybe a version string is the recommended way.

Issue #108, opened 2018-04-16
Use -table instead of -layout for pdftotext?


papascott:
Or make it an option? I find that -table works better at keeping things on the same line when the invoice has columns.

For example, for an invoice like this:
![screenshot_600](https://user-images.githubusercontent.com/8681117/38819991-1ead62fe-419d-11e8-8796-eb7ed91c84d0.png)
I want to match the numbers after "Gesamt". Using -layout it outputs:

```
Reisekostenpauschale5%EUR
EUR17.10
ZwischensummeEUR
MwSt19%EUR359,10
Gesamt68.23

427.33
``` 
while with -table it outputs
```
Reisekostenpauschale5%EUR17.10
ZwischensummeEUR359,10
MwSt19%EUR68.23
GesamtEUR427.33
```

m3nu:
You're right. This should be an option. I suggest to add it to the template. So the template author has full control over how his input data was created.

Small chicken-egg problem of course because we also need to parse the invoice initially to find the keywords. I'll mark it as new feature for now.

rmilecki:
The problem I see is that poppler's `pdftotext` doesn't support `-simple` or `-table`. Support for those layouts was developed in Xpdf after it has been forked by the poppler project.

Today most Linux distributions switched to `poppler` for providing `pdftotext` and similar tools. Ideally support for those extra layouts should be ported from Xpdf to poppler but that seems like a huge task. I created request issue in poppler project to see if there is any idea/interest around that: https://gitlab.freedesktop.org/poppler/poppler/-/issues/1419

PR #118, opened 2018-05-13
adding extract option for pdftotext

duskybomb:
* Adding an option for author to add extract option like `-table` or `-layout`.
 * Refactored `main.py`, `invoice_template.py` and `pdftotext.py` to PEP-8

Solves #108

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ __pycache__/
 # C extensions
 *.so
 
+
+.idea/
 # Distribution / packaging
 .Python
 env/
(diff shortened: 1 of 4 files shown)


duskybomb:
@m3nu Merge this?

m3nu:
We can't add if-conditions like that. To pass arguments to the extract module, there should be a universal way that works the same for all modules.

duskybomb:
But this feature is only for pdftotext module, how can we make it universal when no other module uses such extract layouts?

m3nu:
Other extract options may need special settings as well. If you want to pass those, there needs to be a universal way. Like ffmpeg has for example for filter options.

If you think it's not worth doing it properly we can't add this feature for now.

Issue #114, opened 2018-05-09
Save all fields to json?


msalaj:
Hi, how it is possible to save all fields from template to json file? Currently, only date, amount and desc is written into output file. Program write complete json to screen but it writes only three fields to json file. Thank you

m3nu:
Good idea. We didn't allow custom fields, when the json-export was implemented. Will mark this as new enhancement. Or you can do a PR, if in a hurry.

m3nu:
No. Unit testing means that a single functional unit is tested. So testing extraction and templates at the library level is the right thing to do. The GUI relies on this functional units and adds on them.

duskybomb:
There are two ways to implement this
 * Adding it to template where author specifies fields he wants to dump (I am not in favour of this)
 * Using cli, with an option to dump default fields or all the fields.

m3nu:
Let's always dump all fields. No need to specify or make a template.

duskybomb:
Closing this one.

Issue #119, opened 2018-05-14
Add proper Docstrings to all public functions and classes


m3nu:
To help users get started quickly, let's add some docstrings. I linked to some style guides in the [wiki](https://wiki.debian.org/SummerOfCode2018/Projects/ExtractingDataFromPDFInvoicesAndBillsDetails) that deal with the practice.

duskybomb:
Should I add single line docstring (telling what it does and returns) or a multi-line (giving info about all variables and return values)?

m3nu:
Complex public-facing functions will need multiple lines with all details.

Simple internal helper functions may only need a single line.

m3nu:
In the end, anyone who never used the library needs to know enough about arguments and return values to get started quickly. This is true for invoice2data and even more so for facturx.

duskybomb:
How does this look for extract_data()
```    
"""Function to extract data from e-invoices

:param invoicefile: path of invoice
:param templates: load templates (template name)
:param input_module: input module to use out of input_mapping dict. (default pdftotext)
:return: dict of extracted and matched fields or False if no template matches
"""
```

m3nu:
- The text could use some more detail on what the function does.
- The return desc is fine.
- Parameters also need more details. E.g. if a file point can be passed instead of a string (path).

Like in maths this needs precision because those things make a difference.

duskybomb:
Like this?
```   
 """extract data from e-invoices
     * reads template if no template assigned
     * text is extracted from invoice using input module and stored in extracted_str
     * extracted text is sent for optimization in prepare_input()
     * required fields are matches from templates

    Note: Import all required module when using as a library
        Example: If you want to use pdfminer
            >>> from invoice2data.main import pdfminer
            >>> extract_data("/home/duskybomb/pdf/invoice.pdf", "com.aws-invoice.yaml", pdfminer)

    :param invoicefile: path of invoice (example: "/home/duskybomb/pdf/invoice.pdf")
    :param templates: load templates (template name)
    :param input_module: input module to use out of input_mapping dict. (default pdftotext)

    :var extracted_str: saves extracted text from input module
    :var optimized_str: stores optimized string after passing extracted_str to prepare_input()

    :return: dict of extracted and matched fields, False if no template matches
 """
```

m3nu:
[...]

PR #120, opened 2018-05-14
Export all fields

duskybomb:
Implemented dumping all fields to json and csv.
I need to add it for xml, I want to ask can I use an external library like `dicttoxml`?
#114

### DIFF
--- a/invoice2data/main.py
+++ b/invoice2data/main.py
@@ -74,9 +74,6 @@ def create_parser():
     parser.add_argument('--debug', dest='debug', action='store_true',
                         help='Enable debug information.')
 
-    parser.add_argument('--dump-all-fields', dest='dump_fields', action='store_true',
-                        help='Dumps all fields to output files')
-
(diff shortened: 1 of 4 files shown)


m3nu:
No need for XML.

duskybomb:
Okay no problem, will do it.

m3nu:
If a user takes the effort to define his fields in the template, he probably needs them. Of not, he just ignores them in the export. So having options here adds needless clutter.

m3nu:
To save you time, you should write down the changes you intend to do for each issue before coding them. Then I can make sure you are on the right track.

duskybomb:
Can you merge #120, so that I can then remove conflicts that will appear in #118  and then work on adding docstrings #119 ?

m3nu:
Merging. Please outline the proposed changes for the other features before implementing anything. So we can discuss and go over them before coding anything.

PR #123, opened 2018-05-15
Issue/119: Add docstrings

m3nu:


invoice2data/extract/invoice_template.py
```diff
@@ -31,6 +31,21 @@
 class InvoiceTemplate(OrderedDict):
     """
     Represents single template files that live as .yml files on the disk.
+
+    Methods
+    -------
+    prepare_input(self, extracted_str)
```
m3nu: The `self` argument is passed automatically and shouldn't be included in the docstring.

duskybomb: oh, thanks

duskybomb: Should I remove it or you are?

invoice2data/extract/loader.py
```diff
@@ -35,6 +39,29 @@ def read_templates(folder=None):
...
+
+    Returns
+    -------
+    output : Instance of `InvoiceTemplate`
+        template which match based on keywords
+
+    Examples
+    --------
+
+    >>> read_template("~/home/duskybomb/invoice-templates/")
```
m3nu: This won't work because Python doesn't expand the `~` char. It also doesn't make sense because you already included your home folder.

m3nu: Since just reading the some templates isn't useful, it would be good to show what to do with them later.

invoice2data/extract/invoice_template.py
m3nu: Feel free to update it yourself. I'm working on Factur-X today.

m3nu:
Thanks for the update. Now we wait for those test to pass and then merge it.

duskybomb:
I guess, I will have to review it before merging because you made the PR

duskybomb (approved):

PR #126, opened 2018-05-20
For issue #92

duskybomb:
* Checks for dependencies
    * tesseract
    * imagemagick
* Better message when dependencies not found
Solves #92


invoice2data/input/tesseract.py
```diff
@@ -14,14 +14,21 @@ def to_text(path):
...
+                              stdin=p1.stdout, stdout=subprocess.PIPE)
+        out, err = p2.communicate()
 
-    extracted_str = out
+        extracted_str = out
 
-    return extracted_str
+        return extracted_str
+    else:
+        if spawn.find_executable('tesseract'):
```
m3nu: You misunderstood `find_executable()`. This expression will be `true` if the program is installed. Your later message is thus misleading.

invoice2data/input/tesseract.py
```diff
@@ -14,14 +14,21 @@ def to_text(path):
 
     """
     import subprocess
+    from distutil import spawn
```
m3nu: Here is a typo. The package is called `distutils`. Note the extra "s" at the end.

m3nu:
Thanks. I've moved your checks around a bit. Function is the same, but there are some benefits worth learning:

- fewer if-else functions.
- fewer calls to `find_executable()`
- more logical to have checks at the top. Easier to read later because instructions are followed from top to bottom.

If CI is OK, this is done.

PR #127, opened 2018-05-20
Merge Tests - Issue/8

m3nu:


m3nu:
You had a function to find PDF or JSON files that repeated lots of code. I moved that out to share it between tests. In the future try to follow the DRY principle. Rest seemed OK and we should merge it.

PR #128, opened 2018-05-21
Fixes to PDFminer module and tests.

m3nu:


duskybomb:
Right now they are out right failing on CI nothing to do with templates, right?
I will check the templates.

m3nu:
The current CI failure is due to dependencies. I'll fix those. If you can find the issue with the templates and compare the pdfminer output that will be a bonus.

duskybomb:
The issue with pdfminer is, it doesn't display text in a single line.
Example pdftotext:
```
 Summary
    AWS Service Charges                          $4.11
    Charges                                      $4.11
    Credits                                      $0.00
    Tax *                                        $0.00

 Total for this invoice                          $4.11
```
Pdfminer: It's really difficult to find and match the fields using regex.
```
AWS Service Charges

Charges

Credits

Tax *

Total for this invoice

$4.11

$4.11

$0.00

$0.00

$4.11
```

m3nu:
Understood. Suggest we leave it there. Of people like to use it, they can design their templates for it.

m3nu:
And focus on facturx instead. And the GUI.

Issue #133, opened 2018-05-24
problems with paths or file names containing spaces


flokli:
trying to read a file called "2018-05-09 foo.pdf", invoked via
```
invoice2data --input-reader tesseract 2018-05-09\ foo.pdf --output-format json --output-name foo --debug
```

```
convert: unable to open image `2018-05-09': No such file or directory @ error/blob.c/OpenBlob/2735.
convert: no decode delegate for this image format `' @ error/constitute.c/ReadImage/504.
convert: unable to open image `Call': No such file or directory @ error/blob.c/OpenBlob/2735.
convert: no decode delegate for this image format `' @ error/constitute.c/ReadImage/504.
convert: unable to open image `a': No such file or directory @ error/blob.c/OpenBlob/2735.
convert: no decode delegate for this image format `' @ error/constitute.c/ReadImage/504.
convert: unable to open image `Bike': No such file or directory @ error/blob.c/OpenBlob/2735.
convert: no decode delegate for this image format `' @ error/constitute.c/ReadImage/504.
convert: unable to open image `Rechnung2018-05-09.pdf': No such file or directory @ error/blob.c/OpenBlob/2735.
convert: no images defined `tiff:-' @ error/convert.c/ConvertImageCommand/3258.
Error in pixReadMem: size < 12
Error during processing.
```

m3nu:
Thanks for the report. @duskybomb will look into it. Maybe due to the space that's not passed to Tesseract correctly.

duskybomb:
I can confirm that this error is because of space and we cannot do anything about it, it's the problem with `imagemagick` module. Works fine with `pdftotext`
So just remove white-space from file name and you will be good to go @flokli

m3nu:
Not true, @duskybomb.  While I avoid spaces in filenames, there is generally a way to escape spaces when passing them to the command line.

The issue is with these 2 lines:

```
    convert = "convert -density 350 %s -depth 8 tiff:-" % (path)
    p1 = subprocess.Popen(convert.split(' '), stdout=subprocess.PIPE)
```

This is terrible code because the command is built as string. The correct way is to build it as list and pass it to `subprocess`. E.g. ([from](https://docs.python.org/3/library/subprocess.html) the official docs)

`subprocess.run(["ls", "-l"])`

duskybomb:
Will be using `Popen` instead of `run`, since run is not supported in python2. I tried this:
```
[...]

PR #134, opened 2018-05-24
Remove argument as string for subprocess and use list instead.

duskybomb:
solves #133


flokli:
@duskybomb thanks! will have a look later :-)

We should probably consider to use [wand](http://docs.wand-py.org/en/0.4.4/raphicsmagick) instead of shelling out, seems a lot cleaner too :-)

m3nu:
Definitely. If you can figure out good `wand` options to improve an image for OCR, feel free to add a pull request or open an issue.

PR #159, opened 2018-08-16
Fixes 157, allow non ascii characters

duskybomb:


m3nu:
Thanks for the quick fix.

PR #163, opened 2018-08-31
Add move option

Gsuz:
Added move option and better description to copy.


Gsuz:
Very crude and simple implementation, but works for me :)

m3nu:
Thanks for your commit. It's similar to copying, but I can see cases where you want to see what has been processed before. So let's merge it.

PR #165, opened 2018-09-15
Update invoice_template.py

Gsuz:
Allow banned_keywords in templates


m3nu:
Thanks. Looks good and will likely work. Could you also add a test case and sample template?

And cosmetic stuff: one of the lines is quite long. Should be <80. And you can probably make the code simpler by adding an empty list of banned keywords by default instead of one extra check?

rmilecki:
The same or very similar feature has been implemented in the https://github.com/invoice-x/invoice2data/pull/330

m3nu:
True. Closing this, since it's resolved by #330.

PR #170, opened 2018-11-08
fix unclosed file warning when loading templates

rseabrook:
Fixed an unclosed file warning caused by the fix for #97.


m3nu:
Thanks for the contribution.

PR #173, opened 2018-11-09
Add support for parsing tables

rseabrook:
It is difficult to parse fields with a regex that do not have an identifying format or string on the same line. This happens when an invoice has "table"-oriented fields with multiple column headings on one line followed by their values on the next line.

The ```test/compare/oyo.pdf``` is a good example. The booking ID and payment mode would be hard to parse because they do not have an obvious identifying format that you can target with a regex. 

```
Booking ID                Payment Mode
IBZY2087                  Cash at Hotel
```

You can match those fields with a regex like ```(?P<booking_id>\w+?)\s+(?P<payment_mode>[\w ]+)``` but there are many other matches too.

The proposed ```tables``` plugin borrows heavily from @hbrunn's ```lines``` plugin. It works by defining ```start``` and ```end``` regexes to target a specific section of the invoice. The ```body``` regex attempts to match against the content between start and end and adds any named capture groups as fields in the output. It supports multiple tables per template and does nothing if there are no tables defined in the template. It follows the ```date``` and ```amount``` type conversion conventions like the core extract function.

I can update the readme and tutorial if you think this is a useful addition.

### DIFF
--- a/TUTORIAL.rst
+++ b/TUTORIAL.rst
@@ -149,6 +149,58 @@ and if this one doesn’t match either, this line is ignored. This implies
 that you need to take care that the ``first_line`` regex is the most
 specific one, and ``line`` the least specific.
 
+Tables
+~~~~~
+
+The ``tables`` plugin allows you to parse table-oriented fields that have a row
+of column headers followed by a row of values on the next line. The plugin
+requires a ``start`` and ``end`` regex to identify where the table is located
+in the invoice. The ``body`` regex should contain named capture groups that
+will be added to the fields output. The plugin will attempt to match the
+``body`` regex to the invoice content found between the ``start`` and ``end``
+regexes.
+
+An example invoice that contains table-oriented data may look like:
+
+::
+
+    Guest Name: Sanjay                                                                      Date: 31/12/2017
+
+    Hotel Details                                                   Check In            Check Out       Rooms
+    OYO 4189 Resort Nanganallur,                                    31/12/2017          01/01/2018      1
+    25,Vembuliamman Koil Street,, Pazhavanthangal, Chennai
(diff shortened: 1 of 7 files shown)


m3nu:
This is useful and I keep running into this issue as well.

We should also add more comparison tests/make the existing ones more resilient to e.g. date changes.

Another issue I noticed (now that we approach 5+ extraction methods) is that different extraction tools may give slightly different results in terms of tables and columns. So it would make sense to save the extraction method together with the template and only use that. But that's a separate PR and issue.

m3nu:
Dateparsing is best done but `dateutil`. We just pass the whole string to ist after it's found.

rseabrook:
I updated the Free Fiber comparison test to parse additional fields using the tables plugin. That is the only other test invoice that has a table-formatted section.

In addition, I added the tables plugin to the plugin mapping in ```invoice_template.py``` and updated the tutorial to show how the plugin should be used.

netson:
I just stumbled across the python wrapper for tabula (Java lib):  https://pypi.org/project/tabula-py/
Relying on something like Java might not be desirable, but I've done a few tests with tabula-py on my test system and it works really well. If using a Java lib is unwanted, maybe the logic used by tabula can be implemented in invoice2data? It's open source. I don't know much about Java and I have no clue which method they use for data extraction, but it may be worth looking into! Keep up the good work! 👍

m3nu:
It will be a lot of work to port this. Better integrate it as plugin.

Let's merge this PR here first. Then open an issue for it.

m3nu:
Let's merge this, so a wider audience can work on it. Since it's a plugin, it won't affect other code.

Using and parsing tables is quite essential, so it's good to have some progress in this field.

PR #175, opened 2018-11-23
Include invocice number in renamed pdf to avoid duplicated filenames …

joanjunyent:
…in copy/move mode

Include invocice number in renamed pdf to avoid duplicated filenames in copy/move mode


m3nu:
Thanks. Good change.

PR #179, opened 2018-12-01
add sum_amount_ support to table plugin

shordeaux:


m3nu:
Since it's your first PR, you should go all the way and add a test case and a sample invoice. I'm having a hard time imagining how this works or when.

Without a test case and test PDF, the feature may be broken by someone else in the future. One more reason to add one. (any confidential content can be redacted or replaced in the PDF)

rseabrook:
@m3nu what do you think about refactoring to separate field extraction from type conversion and sum field aggregation? The tables plugin is a new way to extract fields but duplicates the type conversion code from the core lib. If the core lib handled those two tasks as the final step before output, the plugins could take advantage of it. That would simplify the plugins and solve this issue in a cleaner way.

rmilecki:
Any special handling of extracted data **based on fields names** is not a really good idea. It's not flexible enough. It doesn't allow specifying all details on how data should be handled. For the same reason new syntax for `fields` was introduced.

If we need feature as proposed in this pull request we should probably think about syntax like
```
tables:
  - start: Hotel Details\s+Check In\s+Check Out\s+Rooms
    end: Booking ID
    body: (?P<hotel_details>[\S ]+),\s+(?P<date_check_in>\d{1,2}\/\d{1,2}\/\d{4})\s+(?P<date_check_out>\d{1,2}\/\d{1,2}\/\d{4})\s+(?P<amount_rooms>\d+)
  - start: Booking ID\s+Payment Mode
    end: DESCRIPTION
    body: (?P<booking_id>\w+)\s+(?P<payment_method>(?:\w+ ?)*)
    fields:
      amount:
        type: float
        group: sum
```

bosd:
> If we need feature as proposed in this pull request we should probably think about syntax like
> 
> ```
> tables:
>   - start: Hotel Details\s+Check In\s+Check Out\s+Rooms
>     end: Booking ID
>     body: (?P<hotel_details>[\S ]+),\s+(?P<date_check_in>\d{1,2}\/\d{1,2}\/\d{4})\s+(?P<date_check_out>\d{1,2}\/\d{1,2}\/\d{4})\s+(?P<amount_rooms>\d+)
>   - start: Booking ID\s+Payment Mode\s+Amount
>     end: DESCRIPTION
>     body: (?P<booking_id>\w+)\s+(?P<payment_method>(?:\w+ ?)*)\s+(?P<amount>\d+\.\d+)
>     fields:
>       amount:
>         type: float
>         group: sum
> ```

I agree with @rmilecki on the syntax.
Before I did see this thread I made a quick pr to implement the support of coerce types.
https://github.com/invoice-x/invoice2data/pull/498

bosd:
Superseeded by #592

PR #184, opened 2018-12-19
Improve required fields error message

rseabrook:
Log a more detailed error message when required fields are not found. The new message references the list of required fields and the keys in the output dict for quicker debugging.


m3nu:
Seems useful. Thanks.

PR #185, opened 2018-12-19
Add template for mooncard.co

alexis-via:


m3nu:
Can't you merge stuff yourself as admin? For new templates there is no need to have a review or anything I think. :-)

m3nu:
If any setting is blocking you from merging, we should change it.

PR #200, opened 2019-01-27
invoice2data Windows os issues

rafalkasa:
### invoice2data install not working on not English Windows os eg. Polish:

When you try to run `pip install invoice2data` on not English Windows environment below exception occurred because README.rst contains some utf-8 characters:
```
Collecting invoice2data
  Using cached https://files.pythonhosted.org/packages/1a/af/24b281da1ba7fed86a3a936d5ba5dd531d2b33d57202ccda454d07b84fbb/invoice2data-0.3.2.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "C:\Users\RAFALK~2\AppData\Local\Temp\pip-install-sp5t45qw\invoice2data\setup.py", line 14, in <module>
        long_description=open(path.join(path.dirname(__file__), 'README.rst')).read(),
      File "c:\python\lib\encodings\cp1250.py", line 23, in decode
        return codecs.charmap_decode(input,self.errors,decoding_table)[0]
    UnicodeDecodeError: 'charmap' codec can't decode byte 0x98 in position 5608: character maps to <undefined> 
```

### ImageMagick not working on Windows without active shell

`subprocess.Popen` for ImageMagic is not working on Windows environment when the shell is not active. This is because Windows is operating absolutely different on streams compared to the Unix operating systems.

In addition to above Please remember that on Windows exists `C:\Windows\System32\convert.exe` and have the same name as ImageMagick tool `convert`.

### DIFF
--- a/invoice2data/input/tesseract.py
+++ b/invoice2data/input/tesseract.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+import platform
+
+
 def to_text(path):
     """Wraps Tesseract OCR.
 
@@ -19,12 +22,13 @@ def to_text(path):
     # Check for dependencies. Needs Tesseract and Imagemagick installed.
     if not spawn.find_executable('tesseract'):
         raise EnvironmentError('tesseract not installed.')
-    if not spawn.find_executable('convert'):
+    if not spawn.find_executable('convert'):  # Please remember that on Windows exists C:\Windows\System32\convert.exe and have the same name as ImageMagick tool
         raise EnvironmentError('imagemagick not installed.')
 
+    is_shell_active = platform.system() == "Windows"  # ImageMagick on Windows require shell=True for working
     # convert = "convert -density 350 %s -depth 8 tiff:-" % (path)
     convert = ['convert', '-density', '350', path, '-depth', '8', 'png:-']
-    p1 = subprocess.Popen(convert, stdout=subprocess.PIPE)
+    p1 = subprocess.Popen(convert, stdout=subprocess.PIPE, shell=is_shell_active)
 
     tess = ['tesseract', 'stdin', 'stdout']
     p2 = subprocess.Popen(tess, stdin=p1.stdout, stdout=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -11,7 +17,7 @@
     url='https://github.com/m3nu/invoice2data',
     description='Python parser to extract data from pdf invoice',
     license="MIT",
-    long_description=open(path.join(path.dirname(__file__), 'README.rst')).read(),
+    long_description=readme_description(),
```
m3nu: This is probably not even needed any more. The new PyPi version will find README.md/rst/etc anyways.

invoice2data/input/tesseract.py
```diff
@@ -19,12 +22,13 @@ def to_text(path):
...
         raise EnvironmentError('tesseract not installed.')
-    if not spawn.find_executable('convert'):
+    if not spawn.find_executable('convert'):  # Please remember that on Windows exists C:\Windows\System32\convert.exe and have the same name as ImageMagick tool
         raise EnvironmentError('imagemagick not installed.')
 
+    is_shell_active = platform.system() == "Windows"  # ImageMagick on Windows require shell=True for working
     # convert = "convert -density 350 %s -depth 8 tiff:-" % (path)
     convert = ['convert', '-density', '350', path, '-depth', '8', 'png:-']
-    p1 = subprocess.Popen(convert, stdout=subprocess.PIPE)
+    p1 = subprocess.Popen(convert, stdout=subprocess.PIPE, shell=is_shell_active)
```
m3nu: Adding this to all platforms shouldn't cause any issues and would save the extra condition.

setup.py
rafalkasa: @m3nu ok so I will remove this line from `setup.py` and created method readme_description()

invoice2data/input/tesseract.py
rafalkasa: ok, I will change to:

`p1 = subprocess.Popen(convert, stdout=subprocess.PIPE, shell=True)`

invoice2data/input/tesseract.py
```diff
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+import platform
+
+
```
m3nu: unused import?

m3nu:
Thanks. Seeing one last unused import. Then we're good.

m3nu:
Thanks.

PR #203, opened 2019-01-28
Move to src-based layout. Be PEP8-compliant.

m3nu:
Please voice any issues within the next days.

This will allow us to check for PEP8 issues of new PRs and avoid some issues that can come up, when having the package in the root-dir.


m3nu:
It also uses pytest for tests and collects all package settings in the new `setup.cfg` format.

PR #208, opened 2019-02-17
Add template fr.publicationannoncelegale.yml

alexis-via:
Update OVH and Amazon templates for France


m3nu:
Thanks for the update. At some point we should consider adding sample files for each template to enhance the quality.

alexis-via:
Yes, it would be great, but invoices are private documents and it has customer numbers and other information we could be used to contact the supplier on behalf of the real customer. So it's not an easy topic.

PR #211, opened 2019-02-21
fix(xmlOutput): Close file after writing

wollanup:


m3nu:
In my defense: not my code. Using a context would be even better, but this is fine as fix.

Issue #209, opened 2019-02-21
Date format in JSON output is hard coded to %d/%m/%Y


wollanup:
https://github.com/invoice-x/invoice2data/blob/48006efe961584660414c432afbad1ef05a65b3c/src/invoice2data/output/to_json.py#L42

Date format is hard coded to french (maybe more) format `day/month/year`.

Could we add an options like `date_formats_output` or something to have a way to produce more standard format like `%Y-%m-%d` ?

I also noticed that only `date` field is handled, but we ignore fields prefixed with `date` to apply `strftime` on them

m3nu:
Totally agree. Just do a PR to change it.

wollanup:
@m3nu OK, I'm not familiar with Python yet but I'm on it ;)

Please just tell me if this approach looks fine to you :

``` python
def write_to_file(data, path, date_output_format='%d/%m/%Y'):
```
As we have `path` arg, I would add a new arg `date_output_format` in each `write_to_file` functions and pass it in call of `output_module.write_to_file()` from main.py

m3nu:
Yeah. Could be good to add it as optional keyword argument. You want to keep the old default or make it an ISO format?

One small style note: When a string is for formatting, [one](https://stackoverflow.com/a/56190/3983708) would usually use double quotes. `"%d-.."`

wollanup:
OK for double quoting !
Maybe we should keep previous format as default and override on demand to avoid BC ?

m3nu:
Yes. Let's keep the old format.

wollanup:
I'm almost good with code and test, but I introduced a change...
You could call it a Breaking Change or a Bug Fix, you decide ;)

If we consider output JSON of `free_fiber.pdf` file, we had :

```json
{
  "date": "02/07/2015",
  "date_due": "2015-07-05 00:00:00"
}
```
Now, we have this output (without `--output-date-format` option) : 
```json
{
  "date": "02/07/2015",
  "date_due": "05/07/2015"
}
```

And (with `--output-date-format %Y-%m-%d`)
```json
{
  "date": "2015-07-02",
  "date_due": "2015-07-05"
}
```
It sounds better to me because we have only one format for dates.
PR coming soon...

m3nu:
Last option looks best, I think. I'll write a note about it in the release log. Better to fix this now then wait even longer.

I assume most people use it as library anyways, so they get a direct `datetime` object.

bosd:
Fixed by 212

PR #212, opened 2019-02-21
feat(cli): Add option to format dates in output files

wollanup:


wollanup:
See https://github.com/invoice-x/invoice2data/issues/209

m3nu:
Thanks. Will test this tomorrow.

wollanup:
> Thanks. Will test this tomorrow.

Well, I forgot something, I'll fix it before tomorrow...

src/invoice2data/output/to_csv.py
```diff
@@ -48,7 +50,7 @@ def write_to_file(data, path):
             csv_items = []
             for k, v in line.items():
                 # first_row.append(k)
-                if k == 'date':
-                    v = v.strftime('%d/%m/%Y')
+                if k.startswith('date') or k.endswith('date'):
```
m3nu: `if 'date' in k:`

src/invoice2data/main.py
```diff
@@ -117,6 +117,13 @@ def create_parser():
         help='Choose output format. Default: none',
     )
 
+    parser.add_argument(
+        '--output-date-format',
+        dest='output_date_format',
+        default='%d/%m/%Y',
+        help='Choose output date format. Default: %d/%m/%Y',
```
m3nu: Would suggest to use ISO 8601 (2019-02-22), even if it breaks stuff for some people.

src/invoice2data/output/to_csv.py
wollanup: Are you suggering to ignore `startswith()` and `endswith()` and only apply format for `date` key ?

src/invoice2data/main.py
wollanup: OK!

m3nu:
Merged. Sorry for the delay.

PR #213, opened 2019-02-24
Add template for Opal activity statement

squishykid:


m3nu:
Interesting use case.

PR #230, opened 2019-05-25
template: travis-ci.com

loeffel-io:


m3nu:
Thanks for the contribution. You also have a test invoice for it? Can be redacted a bit.

It's not strictly necessary, but more test invoices would be good to make sure it's working after future updates.

m3nu:
Thanks. Merged.

PR #235, opened 2019-06-19
negative amount

daiyyr:
In NZ Watercare's invoices, they use this to indicate a negative total amount (credit balance):
[$ 108.43 cr]
I added IF statement to recognize it as a negative float

PR #240, opened 2019-07-02
negative amount

daiyyr:
1. Dealing with credit balance. Please refer to PR #235  : In NZ Watercare's invoices, they use this to indicate a negative invoice amount (credit balance): $108.43 cr
2. Save source pdf file name as a field 'filename'. Knowing where each CSV line come from will be convenient for debugging. 
3. Fix: unmatching regexp resulting in incorrect CSV line(missing column) when processing multiple PDF


m3nu:
Changes 2 and 3 are great. Thanks for those.

Change 1 is probably too specific to this one invoice format. Have you considered just using an extra field that matches your `cr` expression and then process it some other way?

m3nu:
I'm afraid this is too specific to your own problem. So merging it would increase the code base to maintain without adding much value to most users.

I would recommend to simply use your solution as library between invoice2data's different modules.

PR #241, opened 2019-07-03
Fixed one test and used an env var for the google cloud bucket

EtienneBerube:
The bucket name will be given as an environment variable as the method to_text() is only given one argument when used in extract_data().

The fixed test was using the wrong file extension to compare the results. Therefore, the test would not pass.


m3nu:
I don't think we want to remove the function argument altogether.

EtienneBerube:
Note token. I removed the "default" option of the Google Cloud bucket name as it needs to be changed anyway by a new user (now defaults to None). Now, it can either be an argument, or an environment variable (depends on application). Can now be in a docker.

rakshithrs (approved):
Yes, env var is required for google cloud bucket

rakshithrs (approved):
env var is required for google cloud bucket

PR #244, opened 2019-07-10
Fix crash on passing -h, -he or any substring of help as an argument.

rmNULL:
When any substring of help was passed as an argument in main menu,
the program raised an exception. It happened as argparse uses '%'
internally to format strings, in order to use %, it need to be
escaped with %.


m3nu:
Thanks. Makes sense. Just wondering if this could be achieved with double quotes or some other way that doesn't modify the string.

PR #246, opened 2019-08-07
Change logging to use named loggers everywhere

saulbein:
Allows proper configuration of logging when used as a lib


m3nu:
Thanks. Makes sense.

PR #248, opened 2019-08-28
Chunks Plugin

sam-slater:
Chunks Plugin

more specific multi-lined version of lines.

- "start" and "end" declares the global section where all chunks will be in (Not mandatory. if not stated then uses whole invoice text)

-"chunk_start" and "chunk_end" similar to "start" and "end" but defines the start and end of the single or repeating chunk, e.g start of each meter reading in an invoice (Mandatory)

- "chunk" similar to the "line" key in the lines plugin but allows for multi-line regex expressions (Mandatory)

-"repeat" loops through the content to find every chunk that matches (Not mandatory - defaults to "False")

### DIFF
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -24,7 +24,7 @@ OPTIONS_DEFAULT = {
     "replace": [],  # example: see templates/fr/fr.free.mobile.yml
 }
 
-PLUGIN_MAPPING = {"lines": lines, "tables": tables}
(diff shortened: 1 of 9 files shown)


rhbvkleef:
I wand and/or need this.

rmilecki:
Hi guys,
if you find time to further work on this feature, please feel free to reopen this pull request. As mentioned by @m3nu this needs to be documented. Also with the introduction of parsers we should probably redesign this.
I think we may need to add some kind of content pre-processing syntax that will be reusable for any field.

PR #249, opened 2019-09-04
Update tesseract.py for windows users

goumbalamm:
Update tesseract.py for windows users to avoid conflict with the build-in 'convert' command. Fort imagemagick the command is now 'magick convert' rather than 'convert' only.


m3nu:
Unfortunately the `magick` binary isn't very common or available on many systems. (I tried using Ubuntu 16.04)

I would suggest to make the `convert` binary a function argument to the `tesseract` function with `convert` as default?

PR #253, opened 2019-09-30
Added Lampiris templates

jtommi:
I added a template for the Belgian Energy provider Lampiris.
This is the advance invoice where you pay a fixed amount each month.
Once or twice a year they send a rectification invoice, which is quite different, so it will follow in a new template


jtommi:
I added the regularization invoice template

m3nu:
Thanks. Merged.

PR #265, opened 2019-11-06
Update fr.edf.pme.yml

jonatjano:
almost every field is now parsed,
didn't touch already existant field so it won't change anything for those who where already using the template :smile: 

entry with `consumption` in the name are for the kWh quantity
entry with `unit` in the name are for the price by kWh


m3nu:
Wow. Very comprehensive!

PR #266, opened 2019-11-14
extract_data input_module kwarg accepts strings

jagilley:
I think it makes sense to modify `extract_data` to accept strings as the canonical `input_module` kwarg, as it saves the user from having to instantiate their own `input_module`. (Especially because we have the nice `input_mapping` dict just a few lines earlier!) This change would default the `input_module` kwarg to a string, but still maintain the legacy API because of the `if` clause on line 78 of my fork.


m3nu:
Good to merge this, but docs should be updated too.

PR #267, opened 2019-11-16
Update to match current PDF Miner Version

jischebeck:
The current PDF Miner Version 20191103 (https://github.com/euske/pdfminer/tree/master/pdfminer) doesn't support codec setting.


m3nu:
Thanks.

Issue #283, opened 2020-03-25
Filename of source PDF in output


rschueler:
Is there a way to get the filename of the source PDF in the output?
Because if you parse a whole folder full of PDFs you will not know which record belongs to which file.

rmilecki:
You can simply use the `--output-name` option on `invoice2data`. Make your script read directory files and for each file call `invoice2data` with something like `--output-name $file-parsed`.

PR #285, opened 2020-04-15
Add filename on cli and option to add it on template options

crissto:
Saw #283 and thought it would be a good addition. It's non breaking.

On templates you can set filename: true and it will print it.
I'm not sure about always printing it on CLI mode. Should it be an argument?

Tell me if it's good and I'll add tests.

### DIFF
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -130,7 +130,7 @@ class InvoiceTemplate(OrderedDict):
             return self.parse_date(value)
         assert False, "Unknown type"
 
-    def extract(self, optimized_str):
+    def extract(self, optimized_str, filename=None):
         """
         Given a template file and a string, extract matching data fields.
         """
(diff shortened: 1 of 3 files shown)


m3nu:
Adding this functionality is OK, but you should add it to the right place. `extract()` just works on a string and template. Moving the file is done later [here](https://github.com/invoice-x/invoice2data/blob/7a39049361aa892deac5f81a7b20dea76904b9d9/src/invoice2data/main.py#L222).

crissto:
Right, thanks for the input. 
I've changed it to output the original_file and the moved_file or copied_file.

m3nu:
So the main functionality of this is to preserve the filename in the output folder, if the user specifies the `--copy` or `--move` args?

Are you sure enabling this should be done at the template level? A CLI arg may be better.

Also, could you achieve the same thing by using `--filename-format` and providing the original filename as replacement argument there?

PR #286, opened 2020-04-22
Allow setting custom lines join separator

rmilecki:
```
Replace hardcoded "\n" with plugin option. It allows specifying custom
separator in template and e.g. joining lines with a space.

```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -157,6 +157,7 @@ Example:
         first_line: ^    (?P<description>\w+.*)\$(?P<price_unit>\d+\.\d+)
         line: (.*)\$(\d+\.\d+)
         last_line: VAT \*\*
+        lines_join: " "
 
 Development
 -----------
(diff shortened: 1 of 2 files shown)


m3nu:
What do you think @hbrunn?

hbrunn:
looks good to me.

my yaml is quite weak, so I'd suggest to show in the readme how to make this a `\r` or similar

rmilecki:
I'm still thinking about this change myself. I proposed this change as I don't reason for having `\n` in invoice product/service name.

Now I'm wondering if there are any product/service names that require specified line breaks at all? I think line breaks always appear due to text wrapping.

Why anyone would ever need `\n`? Maybe we could change `lines` plugin behavior by default?

On the other hand, what if I want invoice product/service names from company Foo parsed with spaces and someone else with line breaks (why?). Maybe it's indeed up to `invoice2data` user to replace `\n` with ` `?

rmilecki:
This was a stupid idea, tpl files should be generic, it's up to application to transform data if needed.

PR #287, opened 2020-04-23
Export all data to XML format

rmilecki:
```
Export everything (including lines) instead of 4 predefined fields so
XML can be fully used.

Before:
<?xml version="1.0" ?>
<data>
  <item id="1">
    <date>2020-04-15</date>
    <desc>Invoice from Foo company</desc>
    <currency>PLN</currency>
    <amount>123.45</amount>
  </item>
</data>

After:
<?xml version="1.0" ?>
<data>
  <item id="1">
    <issuer>Foo company</issuer>
    <amount>123.45</amount>
    <date>2020-04-15</date>
    <invoice_number>1234</invoice_number>
    <currency>PLN</currency>
    <lines>
      <item>
        <pos>1</pos>
        <name>Bread</name>
        <qty>5</qty>
      </item>
      <item>
        <pos>2</pos>
        <name>Milk</name>
        <qty>10</qty>
      </item>
    </lines>
    <desc>Invoice from Foo company</desc>
  </item>
</data>

```

### DIFF
--- a/src/invoice2data/output/to_xml.py
+++ b/src/invoice2data/output/to_xml.py
@@ -1,4 +1,5 @@
 import xml.etree.ElementTree as ET
+import datetime
 from xml.dom import minidom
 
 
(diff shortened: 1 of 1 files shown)


rmilecki:
Ping :)

rmilecki:
> Sensible change, but will break backwards compat.

It seems to me that all "old" XML nodes are preserved and my change only adds new ones. Adding new XML nodes should not break any app unless it horribly badly designed.

m3nu:
You're right. Let's add it then. Thanks for the contribution! 🙏

PR #291, opened 2020-06-11
Completed the XML File Export

sudeepjd:
3 updates

1. The remove_whitespace removed all the whitespaces merging the words together... Updated this to keep remove all the whitespaces, except 1, this will allow the use of (\w) regex

2. Removed the print statements in to_json.py as they were coming up in the command line.

3. XML export only had few items being exported, update the to_xml.py to add in all the line in a structured format. The tags become the keys and the text as the values.. The invoice lines have <lines> and <line> tags. This is based on the list datatype. Have not tested for tables though.


sudeepjd:
Realised there was already a pending request for the same function...

PR #304, opened 2020-09-30
Mostly using pip3 now, adapt install instructions

StephanEggermont:


m3nu:
Most `pip3`s are just `pip` now. Not sure we need to add this.

StephanEggermont:
I wouldn't know. Following the instructions (Mac 10.14.6 with Xcode 11.2.1 installed) there is no pip. Sounds to me like the instructions only work if you have other things installed already.

m3nu:
With Python 2 support ending, most Pythons will be Python 3 in the future. So `pip` and `pip3` will be the same. So no need to clutter our docs and increase mental load for readers. Thanks for the idea anyways!

Issue #307, opened 2020-10-27
Idea: allow "fields" entries to use extended syntax with associative arrays


rmilecki:
Right now `fields` allows specifying:
1. Regex to use for extracting field data (e.g. `foo: Order (\d+)`)
2. Static value (e.g. `static_bar: Baz`)
3. Sums (e.g. `sum_amount_qux: ...`)

I find current syntax (described above) is a bit loose, limiting and not allowing some further clean development. For example you can't have field starting with `static` using regex (e.g. `static_ip_address: (\d+\.\d+\.\d+\.\d+)`). You can't add new feature like converting extracted value (where to put info how to proceed?).

I'd like to suggest extending `fields` format to support syntax like:
```
fields:
  foo:
    regex: Order (\d+)
  bar:
    static: 'Baz'
  qux:
    sum: ...
```
This would allow specifying field extraction explicitly.

It would also allow adding new features cleanly, things like:
```
fields:
  phone:
    regex: Phone ([\d-]+)
    sub:
      pattern: '-'
      replace: ''
```
Any opinions on this idea?

rmilecki:
Implemented by the https://github.com/invoice-x/invoice2data/pull/308

Issue #296, opened 2020-07-16
Is there a way to parse multiple tables using "lines"


suprateekn:
I am having 2 tables in a single page of the pdf and want to extract data from both of them. Can't use the "tables" plugin as can have only the first line of data using that.

As far as I can understand the "lines" block is treated as a dictionary. So multiple table parsing using "lines" is not possible. Is there an alternative?

rmilecki:
I can implement this but I need `invoice2data` maintainers to help me to decide on YAML syntax for it. I have two suggestions.
@m3nu: can you comment on below ideas, please?

### Use extended `fields` syntax
In the #307 I suggested each `fields` entry to be an associative array. That would allow very clean support for requested feature (without breaking backward compatibility), consider:
```
fields:
  foo:
    static: 'Lorem ipsum'
  items:
    plugin: lines
    settings:
      start: ...
      end: ...
      line: ...
  rates:
    plugin: lines
    settings:
      start: ...
      end: ...
      line: ...
```
That would require rewriting plugins API a bit which I can easily handle. The only problem I see is that `table` plugin wouldn't match that design. It's because `table` plugin parses (returns) multiple fields. It means we may need two APIs for plugins then (which is not a problem for me - I'm just making it clear).

### Extend existing `lines` syntax
Current syntax for `lines` looks like this:
```
lines:
  start: ...
  end: ...
  line: ...
```
We could extend it to support following (without breaking backward compatibility):
```
lines:
  - items:
      start: ...
      end: ...
      line: ...
  - rates:
      start: ...
      end: ...
      line: ...
```

suprateekn:
Instead of this I was thinking if the lines section could be made like the tables section. Where we could give multiple entries. That would be really helpful

rmilecki:
> Instead of this I was thinking if the lines section could be made like the tables section. Where we could give multiple entries. That would be really helpful

The difference between above "Extend existing lines syntax" and `tables` plugin syntax is the former having every array entry named (`items` and `rates`). We can't have pure `tables`-like syntax instead. It's because:
1. `lines` plugin returns array that has to be assigned to some field
2. `tables` plugin assigns to few fields depending on used `body`

Unless I misunderstood you. If so, please provide some syntax example, so it's clear what you mean.

m3nu:
> Use extended fields syntax

[...]

PR #308, opened 2020-10-30
Extend fields syntax to support associative arrays & add 2 initial parsers

rmilecki:
This pull request implements ideas and features described / requested in the:
- #296
- #307

This adds support for following syntax:
```
fields:
  friendly_name:
    parser: static
    value: Amazon
  rates:
    parser: lines
    start: ...
    end: ...
    line: ...
```

I took care of properly describing every commit so let me just copy & paste their messages here:
```
parsers: translate lines plugin into a parser

This allows reusing code parsing lines in multiple fields. It doesn't
break backward compatibility as the old lines plugin syntax is still
supported.

Syntax example:

fields:
  example:
    parser: lines
    start: ...
    end: ...
    line: ...
```
```
parsers: add static parser

It's a trivial pseudo-parser that matches a behaviour of fields defined
using static_foo name. It simply returns content of the "value".

Syntax example:

fields:
  example:
    parser: static
    value: Lorem ipsum
```
```
add parsers concept for fields defined as YAML associative arrays

So far fields could be defined only as a name + regex pair. All special
cases were handled in a bit hacky way - by using properly formatter
field name (e.g. "static_foo" or "sum_amount_foo").

This change allows defining field in YAML template as an associative
array (Python dictionary). It's a cleaner way, results in more scalable
design and allows specifying more per-field details if needed.

The new syntax requires specifing parser by its name. Name "parser" was
used to distinguish it from plugins. It should be also a bit more
meaningful as plugins are often considered some external code.

Syntax example:

fields:
  example:
    parser: foo
    data: bar
```

### DIFF
--- a/TUTORIAL.rst
+++ b/TUTORIAL.rst
@@ -53,10 +53,18 @@ Fields
 
 All the regex ``fields`` you need extracted. Required fields are
 ``amount``, ``date``, ``invoice_number``. It’s up to you, if you need
-more fields extracted. Each field has one or more regex with one regex
-capturing group. It’s not required to put add the whole regex to the
-capturing group. Often we use keywords and only capture part of the
-match (e.g. the amount).
+more fields extracted. Each field can be defined as:
+
+-  an **associative array** with ``parser`` specifying parsing method
+-  a single regex with one capturing group
+-  an array of regexes
+
+The first method is preferred. It was introduced to make templates
+syntax cleaner and more flexible. It aims to replace old methods.
+
+It’s not required to put add the whole regex to the capturing group.
+Often we use keywords and only capture part of the match (e.g. the
+amount).
 
 You will need to understand regular expressions to find the right
 values. If you didn’t need them in your life up until now (lucky you),
@@ -66,6 +74,9 @@ here <http://www.regexr.com/>`__. We use `Python’s regex
 engine <https://docs.python.org/2/library/re.html>`__. It won’t matter
 for the simple expressions we need, but sometimes there are subtle
 differences when e.g. coming from Perl.
+
+Legacy regexes
+~~~~~~~~~~~~~~
 
 For non-text fields, the name of the field is important:
 
--- a/TUTORIAL.rst
+++ b/TUTORIAL.rst
@@ -75,6 +75,20 @@ engine <https://docs.python.org/2/library/re.html>`__. It won’t matter
 for the simple expressions we need, but sometimes there are subtle
 differences when e.g. coming from Perl.
 
+Parser ``static``
+~~~~~~~~~~~~~~~~~
+
+This pseudo-parser sets field with a content of ``value`` field.
+
+Example:
+
+::
+
+    fields:
+      friendly_name:
+        parser: static
+        value: Amazon
+
 Legacy regexes
 ~~~~~~~~~~~~~~
 
--- a/TUTORIAL.rst
+++ b/TUTORIAL.rst
@@ -75,6 +75,44 @@ engine <https://docs.python.org/2/library/re.html>`__. It won’t matter
 for the simple expressions we need, but sometimes there are subtle
 differences when e.g. coming from Perl.
 
+Parser ``regex``
+~~~~~~~~~~~~~~~~
+
+It's the basic parser that allows parsing content using regexes. The
+only required property is ``regex`` that has to contain one or multiple
+(specified using array) regexes.
+
+By default ``regex`` parser removes all duplicated matches. It results a
+single value or an array (depending an amount of unique matches found).
+
+Optional properties:
+
+-  ``type`` (if present must be one of: ``int``, ``float``, ``date``) -
+   results in parsing every matched value to a specified type
+-  ``group`` (if present must be ``sum``) - results in grouping all
+   matched values using specified method
+
+Example for ``regex``:
+
+::
+
+    fields:
+      amount:
+        parser: regex
+        regex: Total:\s+(\d+\.\d+) EUR
+        type: float
+      date:
+        parser: regex
+        regex: Issued on:\s+(\d{4}-\d{2}-\d{2})
+        type: date
+      advance:
+        parser: regex
+        regex:
+          - Advance payment:\s+(\d+\.\d+)
+          - Paid in advance:\s+(\d+\.\d+)
+        type: float
+        group: sum
+
 Parser ``static``
 ~~~~~~~~~~~~~~~~~
 
--- a/TUTORIAL.rst
+++ b/TUTORIAL.rst
@@ -89,6 +89,27 @@ Example:
         parser: static
         value: Amazon
 
+Parser ``lines``
+~~~~~~~~~~~~~~~~
+
+This parser allows parsing selected invoice section as a set of lines
+sharing some pattern. Those can be e.g. invoice items (good or services)
+or VAT rates.
+
+It replaces ``lines`` plugin and should be preferred over it. It allows
+reusing in multiple ``fields``.
+
+Example for ``fields``:
+
+::
+
+    fields:
+      lines:
+        parser: lines
+        start: Item\s+Discount\s+Price$
+        end: \s+Total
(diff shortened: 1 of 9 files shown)


m3nu:
Hi and thanks for the contribution! Before merging, wanted to ask the following:

### Plugins vs. parsers
We need to clearly distinguish `plugins` and `parsers`. From what I see, `parsers` process the `fields` section in a template and returns a single field, while `plugins` can do whatever, including returning arbitrary fields, calling external tools, etc. (e.g. [here](https://github.com/invoice-x/invoice2data/blob/d2984e09e211449b2798e9d2338b764f1a02847a/src/invoice2data/extract/templates/com/com.apple.yml)). This is fine to give maximum flexibility. So `parsers` would be a core part of invoice2data, while plugins are more loosely associated.

The `lines` module looks more like a plugin, as it can return multiple fields. So maybe it should stay a plugin, rather than parser? (Unless we change the definition of parser vs. plugin.) If we make lines a `parser` and a user needs multiple fields, the line config would be repeated, right? So the plugin option works better here.

While we need a better separation of plugin vs. parser, I'm very much in favor of the YAML config extension for fields to make the parser more explicit, as you did. I do suggest to make the "regex" parser explicit as well and move related code into the `parsers` package?

We'll also need to define supported template options somewhere in the docs:

```
fields:
  friendly_name:
    parser: static
    value: Amazon
  rates:
    parser: regex
    pattern: XYZ\d+
```

If you think `lines` should be both, plugin and parser, we should at least make a comment on that in the code, clearly explain the difference and possibly share some code to avoid duplication. I do tend to keeping lines as plugin only.

Since we expand the syntax, can you think of other examples of parsers that could be added in the future? Just to avoid expanding the syntax and then always using regex anyways :-)

rmilecki:
> We need to clearly distinguish `plugins` and `parsers`.

I'm sorry I didn't make it clear in the first place. You're right of course.
- Parsers can fill a single output entry only (they are bounded with `fields` entry they were specified in). As (poorly) documented, they can return "e.g. number, string, array"
- Plugins can do "anything" with the output (e.g. set multiple entries like `table` does)

Parsers fit templates design better thanks to being utilized in the `fields`. That means keeping parsing details in one place (`fields`) which I find a better & cleaner design. That's why I believe we should **prefer** parsers over plugins where possible.

So a quick look at `lines`:
1. it always returns an array
2. we prefer parsers over plugins

that's why I'd like to see `lines` being used as **parser**. It also solves the problem of reusing `lines` for multiple `fields` (#296) without extra `lines` specific. I see it as a win-win.

> I do suggest to make the "regex" parser explicit as well and move related code into the parsers package?

This is **definitely** on my list, probably as the next task. I'll need to think about proper design (so we can handle sums, amounts, types properly). Once I design it, I'll share it for public discussion.

> We'll also need to define supported template options somewhere in the docs:

I intentionally didn't update documentation in my commits as I wanted to handle it right before the next release. I didn't want to document features that are present in the `master` branch only. Is that OK?

> If you think lines should be both, plugin and parser, we should at least make a comment on that in the code

I want to keep it working as plugin for backward compatibility. I don't want to break people templates. Luckily it takes only 3 lines of code. We may announce we're planning to drop `lines` **plugin**, ask people to switch to `lines` **parser** and drop it in a year or so.

> Since we expand the syntax, can you think of other examples of parsers that could be added in the future? Just to avoid expanding the syntax and then always using regex anyways :-)

I don't have any **new** ideas right now. I think we may have some discussion while adding `regex` parser (e.g. should a separated parser be used for summing?). In any case I believe that this new syntax should be flexible enough to handle further cases. The only limitation is that parser can fill only a single output entry (return a single value).

rmilecki:
@m3nu: I've updated my branch editing commits a bit:
1. I updated the first commit message so it explains idea behind parsers
2. I worked on basic explanation in the `parsers/__interface__.py` and `plugins/__interface__.py`

m3nu:
I think the regex, sum, amount and date options should be explicit YAML options with parser as well. Then keep minimal code in `invoice_template.py` for backwards-compat. Else we have some code in YAML and some code in the old `invoice_template.py`.

Last available YAML parsers should be mentioned in the docs.

m3nu:
> Do you mean I should document them now, as part of this pull request? I can do that, no problem, just asking for clarification.

If we add the changes without documentation nobody will now about them. Unless you want to keep them secret and document later?

rmilecki:
> I think the regex, sum, amount and date options should be explicit YAML options with parser as well. Then keep minimal code in `invoice_template.py` for backwards-compat.

I fully agree. I'm going to work on that next.

> Last available YAML parsers should be mentioned in the docs.

Do you mean I should document them **now**, as part of this pull request? I can do that, no problem, just asking for clarifications.

rmilecki:
@m3nu: I've just updated my branch with relevant documentation changes. I'm not native English so it may require some tweaking.

Do you see any other blockers for this Pull Request as this point?

Ideally I'd like to see it merged and work on other parsers in about a week.

m3nu:
I think you're on a good track. Docs, etc are good.

I do think we should move the code below this [line](https://github.com/invoice-x/invoice2data/blob/cfe11dda08c6331391489722ac092a176fa0ff6a/src/invoice2data/extract/invoice_template.py#L179) into a regex parser module? So `invoice_template.py` is working on a higher level, calling different parsers. This regex parser would include all the magic we have for sum, date, etc and would be the default if a field only has a string (=single regex) or list (=multiple regex). If it's a dict, the new explicit interface applies.

Tests and Flake8 should also pass before merging.

Sorry for all the input and changes. It's just a big change, which will define the project structure for many years. So better get it right and consider everything.

rmilecki:
Oh, we probably misunderstood each other a bit. My idea was to "add 2 initial parsers" in this pull request and work on regex parser "as the next task" (as a new pull request). I think you expected regex parser to be added in this pull request from the beginning.

I hope we're on the same page now, let me work on the regex parser and failing tests :) I'll try to get this done this week.

rmilecki:
I've added `regex` parser, documented it and modified legacy syntax code to use it. I'm really happy how clean the code looks now.

I believe the only remaining issue is complaining `flake8`:
```
src/invoice2data/extract/parsers/__init__.py:3:1: F401 '.lines' imported but unused
src/invoice2data/extract/parsers/__init__.py:4:1: F401 '.regex' imported but unused
src/invoice2data/extract/parsers/__init__.py:5:1: F401 '.static' imported but unused
```

I added those imports to the `parsers/__init__.py` to let `invoice_template.py` use parsers cleanly as follows:
```
from . import parsers

PARSERS_MAPPING = {"lines": parsers.lines, "regex": parsers.regex, "static": parsers.static}
```
(but `flake8` doesn't like it apparently).

The only different solution I'm aware of is doing:
```
from .parsers import lines as parsers_lines
from .parsers import regex as parsers_regex
from .parsers import static as parsers_static
```
(it should work for `flake8` but I find it a bit less elegant)

Info I found: https://stackoverflow.com/questions/8893209/relative-import-of-submodule

@m3nu: can you review the `regex` parser and comment on my `import`s issue, please?

m3nu:
Yeah, Flake8 probably doesn't see the "usage" in the dict mapping. No problem to just [override](https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors) this with `# noqa F401`

Definitely a good rewrite. It will be quite easy to add new parsers in the future.

Let's give people a few days to comment on this and then merge it. Since the API changed, we'll bump the version for this. Good job!

rmilecki:
Thank you @m3nu!

If I read CircleCI correctly, two tests didn't pass because `python3.7` is not available anymore and those tests were skipped.

m3nu (approved):

m3nu:
Merged. Thanks for the patience. 🤝

m3nu:
> If I read CircleCI correctly, two tests didn't pass because python3.7 is not available anymore and those tests were skipped.

No worries about this. Would need slight adjustment to make Python 3.7 available on the CI.

Eventually we should switch to Github actions though. Circle and Travis aren't as widely used any more.

PR #322, opened 2021-02-12
Update example template

keeeenion:
As the parser formats and then removes "static_" prefix and only returns "vat" part in the result object this example will never succeed as "static_vat" never exists in the result.


m3nu (approved):

m3nu:
Thanks! Good catch.

Issue #323, opened 2021-02-25
Problems using gvision


juanluisrosaramos:
Google has (once again) changed its code to consume its Cloud Vision APIs.  Mostly, my invoices are photos without text to copy.  So I need an OCR and google vision has much better results than tesseract.  So I will update your code I will love to contribute to your software afterwards.

The error that I have is "AttributeError: module 'google.cloud.vision' has no attribute 'types'" so I need to update the code to the last version.

m3nu:
How annoying. Feel free to add a pull request to fix the issue. After that we can make a small release.

juanluisrosaramos:
#324

bosd:
This issue and PR had been open for quite soms while. Propably something we need to work on.


@juanluisrosaramos is your code in the PR still actual?

PR #324, opened 2021-02-26
#323 Update gvision.py to changes in gcloud vision library

juanluisrosaramos:
Updated to google.cloud.vision recent changes (no more type)
Updated new json response name in gcloud bucket
Updated to last code modifications in https://cloud.google.com/vision/docs/pdf?hl=en
Added the capacity to extract text from all the pdf pages
Suggestions: abstract json response file name and parametrize bucket name


m3nu:
Does it make sense to add a test here?

bosd:
superseeded by #576

PR #305, opened 2020-10-18
Area Plugin Support 

kavinsharma:
An invoice2data `area `plugin helps in extracting text on the basis of area coordinates utilizing [pdf2text area cropping option](https://prnt.sc/v1uvhl). An area plugin is customization to invoice2data to define area cropping with coordinates. Coordinates defined for the template can vary from pdf to pdf.
You just have to add a normal template containing the YAML file in which there are different plugins for fields and tables and you just have to add an area plugin and it works on every pdf.
Just write the field of multiple lines you want to extract and give the coordinates of that field that is (`x=? y=? r=? H=? W=?`)
```
x = x coordinate
y = y coordinate
H= height of the area
W= width of the area 
R= resolution of invoice
```
**Area Plugin Options:** 
Name: field name to map with extracted text
Area: takes dict as input for cropping pdf area and extract text
Regex: Optional parameter, used for further extracting text from the cropped area.
[Sample Invoice ](https://prnt.sc/v1uyl6)

Here is a sample of an invoice template including the area plugin that helps you extract multiple lines.
```
# -*- coding: utf-8 -*-
issuer: The XYZ Company
keywords:
- The XYZ Company
- US Supplier 123
fields:
	amount: TOTAL:\s+(\d+,\d+\.\d\d)
	date: Invoice Date:\s+(\d{1,2}\/\d{1,2}\/\d{4})
	delivery_date: Delivery Date:\s+(\d{1,2}\/\d{1,2}\/\d{4})
	invoice_number: INVOICE:\s+(\w{3}\d{1,8})
	sales_order : Sales Order:\s+(\w{2}\d+)
tables:
	-   start: Line\s+Product\s+Description\s+Quantity
    	end: Prices
    	body: (?P<Line>^\d{2})\s+(?P<Product>\w{2}\-\w{2}\-\w+\-\w+)\s+(?P<Description>\w+\s\w+\s\w+\-\w+)\s+(?P<Quantity>\d+)
area:
    -  name: "Address"
       area: {x: 115,y: 124,r: 300,W: 412, H: 326}
       regex: \w+
     - name: "Bill_to"
       area: {x: 930,y: 416,r: 300,W: 310, H: 294}

options:
	remove_whitespace: false
	currency: USD
	date_formats:
    	- '%d/%m/%Y'
	languages:
    	- en
decimal_separator: '.'

```
Output :
```
[...]

PR #333, opened 2021-03-17
Added area support for fields

RossK1:
Fields specifying a parser now also accept an "area" argument. This allows one to specify the page(s) and coordinates of the area to be extracted and then applies the specified parser to the results as normal :)
Currently this only works for pdftotext but could be expanded to other converters fairly easily.
Incorporated changes into all existing functions where possible to allow for ease of future maintainability.
Tutorial was updated to show new option and its syntax.

Credit for the idea and base code goes to @kavinsharma in issue #305

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -58,7 +58,12 @@ All the regex `fields` you need extracted. Required fields are `amount`,
 `date`, `invoice_number`. It's up to you, if you need more fields
 extracted. Each field can be defined as:
 
-- an **associative array** with `parser` specifying parsing method
+- an **associative array** with 
+`parser` (required) specifying parsing method and 
+`area` (optional) specifying the region of the pdf to search. 
+This takes the following arguments: `f` (first page), `l` (last page), `x` (top-left x-coord), `y` (top-left y-coord), 
+`r` (resolution), `W` (width in pixels) and `H` (height in pixels). When setting your region, ensure the resolution in your 
+image editor matches the resolution specified for `r` in this option. If not, it will not line up properly.
 - a single regex with one capturing group
 - an array of regexes
 
@@ -103,6 +108,7 @@ Example for `regex`:
         type: float
       date:
         parser: regex
+        area: {f: 1, l: 1, x: 110, y: 50, r: 300, W: 100, H: 200}
         regex: Issued on:\s+(\d{4}-\d{2}-\d{2})
         type: date
       advance:
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -214,7 +214,7 @@ class InvoiceTemplate(OrderedDict):
                     # e.g. "parser: regex" requires "regex: [pattern]"
                     if v["parser"] in PARSERS_MAPPING:
                         parser = PARSERS_MAPPING[v["parser"]]
-                        value = parser.parse(self, v, optimized_str)
+                        value = parser.parse(self, v, optimized_str_for_parser)
                         if value is not None:
                             output[k] = value
                         else:
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -11,6 +11,9 @@ import logging
 from collections import OrderedDict
 from . import parsers
 from .plugins import lines, tables
+# Area extraction is currently only added for pdftotext
+from ..input import pdftotext
+from typing import Optional
 
 logger = logging.getLogger(__name__)
 
@@ -71,7 +74,7 @@ class InvoiceTemplate(OrderedDict):
         Input raw string and do transformations, as set in template file.
         """
 
-        # Remove withspace
+        # Remove whitespace
         if self.options["remove_whitespace"]:
             optimized_str = re.sub(" +", "", extracted_str)
         else:
@@ -81,11 +84,11 @@ class InvoiceTemplate(OrderedDict):
         if self.options["remove_accents"]:
             optimized_str = unidecode(optimized_str)
 
-        # convert to lower case
+        # Convert to lower case
         if self.options["lowercase"]:
             optimized_str = optimized_str.lower()
 
-        # specific replace
+        # Specific replace
         for replace in self.options["replace"]:
             assert len(replace) == 2, "A replace should be a list of 2 items"
             optimized_str = optimized_str.replace(replace[0], replace[1])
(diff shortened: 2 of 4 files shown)


RossK1:
@kavinsharma, I hope this does justice to your original idea :) Based on some of the comments on #305 I switched it into the parsers to allow people to extract areas and still be able to parse them using the other functionalities.
@m3nu, let me know what you think :)

m3nu:
This should probably be a parser, if it only finds a single value in a specific area.

Difference between parser and plugin: https://github.com/invoice-x/invoice2data/blob/master/src/invoice2data/extract/plugins/__interface__.py#L4

m3nu:
The code in this PR is not a parser currently, since it's not **one file** in the `parsers` [package](https://github.com/invoice-x/invoice2data/tree/master/src/invoice2data/extract/parsers).

We added parsers and plugins to avoid cluttering the rest of the code too much.

Hope you guys can work together on this. Personally, I'd build on #305, since it's more neatly separated. I'll add some comments there.

RossK1:
@m3nu My idea with having it as an argument rather than a plugin or separate parser is that it could be used in conjunction with any of the other parsers. E.g. if you need to pull out lines from a specific region, you can specify that area and the lines parser. If you need a single value, specify the area and the regex parser. This way any future parsers inherit the functionality automatically as the area extraction is checked before the parser is called :)
It is also completely optional, so any existing templates without the area specified work as normal and if you want to search the entire document, you can just leave it out.
Thoughts?

erkin98:
when it'll merge?

> @m3nu My idea with having it as an argument rather than a plugin or separate parser is that it could be used in conjunction with any of the other parsers. E.g. if you need to pull out lines from a specific region, you can specify that area and the lines parser. If you need a single value, specify the area and the regex parser. This way any future parsers inherit the functionality automatically as the area extraction is checked before the parser is called :)
> It is also completely optional, so any existing templates without the area specified work as normal and if you want to search the entire document, you can just leave it out.
> Thoughts?

kavinsharma:
ok sure @m3nu , we will look at your comments and fix it up. @RossK1 I hope you will be fine in contributing to my PR #305  :)

erkin98 (approved):

RossK1:
> ok sure @m3nu , we will look at your comments and fix it up. @RossK1 I hope you will be fine in contributing to my PR #305 :)

For sure! If you need me to test or check something in it, let me know 😄 
Even if #305 is implemented as a plugin, I think the functionality in #333 is still important to have as it allows the parsers to (optionally) be directed to act on a very specific area of the invoice. I have found this to be required when dealing with invoices where multiple pieces of information are presented beside each other on the page, but move in vertical relation to each other, causing pdftotext to return slightly mis-aligned lines for each pdf. Being able to clip out the part you want and still run regex with sums or lines on it is invaluable.

RossK1:
@m3nu, what do you think? Would this be a useful feature? You'll see from the extra code that if the "area" parameter is passed in a field, it simply re-runs pdftotext for the region specified and uses the extracted text for that specific parser. Other fields and other plugins are unaffected and yet this functionality can benefit any current and future parsers meaning contributors can focus on the parser instead of how to extract an area of the page.

erkin98:
> @m3nu, what do you think? Would this be a useful feature? You'll see from the extra code that if the "area" parameter is passed in a field, it simply re-runs pdftotext for the region specified and uses the extracted text for that specific parser. Other fields and other plugins are unaffected and yet this functionality can benefit any current and future parsers meaning contributors can focus on the parser instead of how to extract an area of the page.

is it available for usage?if is how can we use ?

RossK1:
@m3nu any update on reviewing and merging this feature?

m3nu:
Didn't you want to make this as parser? It also needs a rebase.

RossK1:
If we make it a separate parser, it loses the ability to leverage the other parsers. Being an optional input to any field, it simply narrows down the content being sent to any other parser. This way the other parsers do not need their own area support, it is handled up-front.

m3nu:
True, that's an argument for adding it on top and I tend to agree. Let me consider it again tomorrow.

RossK1:
Yeah, as much as I would like to avoid cluttering up the code, this seems like a more elegant way of doing it without having every parser have to call some other area parser first.

RossK1:
@m3nu . Any updates on this PR?

Thanks!

pranavshah01:
I have this same requirement. Any update on this PR?

m3nu:
I would have preferred this as standalone parser, but see the tight linkage with `pdftotext`. So the solution you chose makes sense.

Please just add a few test cases using the existing (or new) sample files to make sure the feature keeps working in the future.

Then we can merge and release this feature.

m3nu:
Yes, some tests please. So we will know if this feature breaks in the future.

m3nu:
And please close any competing/superseded area plugin PRs. We had a few of them.

alexx-avdeev:
is there any update on this?

m3nu:
- [ ] Rebase and resolve conflicts
- [ ] Add tests and test files (or reuse existing files)
- [ ] Close competing or superseded PRs

RossK1:
For the tests, would it be sufficient to simply reuse the Amazon one? Update the template file (src/invoice2data/extract/templates/com/com.amazon.aws.yml) to include an area and then update the expected result file (tests/compare/AmazonWebServices.json) to reflect the area to be extracted? The test_content_json in tests/test_cli.py should pick up and compare the changes, right?
Would this one test be enough or would you like to see similar extractions for multiple files?
Thanks :)

m3nu:
Sure. No problem to reuse an existing template if you don't want to share your own documents.

The test should run the new code. If you reuse a document, please add a new test and comparison output, so the old test is still there.

Logic-Bits:
Hi!
Any updates on this?

Thanks

bosd:
> - [ ] Rebase and resolve conflicts
> - [ ] Add tests and test files (or reuse existing files)
> - [ ] Close competing or superseded PRs

@RossK1 This PR has been sitting here for a while. Would be great if it could be included in the next release. Could you attend to the outstanding tasks?

bosd:
@alexis-via  Great to see you back here!
@m3nu  Since there are approving reviews, can we still include this in the 0.4.0 release?

alexis-via:
As the author of this PR didn't answer messages for several months, I rebased and added a test in this new PR: https://github.com/invoice-x/invoice2data/pull/438

bosd:
Ok, this one can be closed now.
Thanks guys, for all your efforts!!

PR #334, opened 2021-03-18
Modified to_csv documentation to clarify input data type

RossK1:
CORRECTION: It is simply a documentation issue. The original doc indicates a dictionary being passed for the data. It is, in fact, a list of dictionaries that is required.

I was unable to get the previous to_csv.py file to work. The data I was sending to it consisted of single fields and lines. It was trying to find keys in all values, regardless of whether they were dictionaries or not.

It appears that any lines-type regex will return a list of dictionaries, so I modified the code to now check if the field is a list of dictionaries first. Only then does it start looking for keys and values in each element.

All values where the name starts or ends with "date" is still attempted to be parsed as a date format whether it is on it's own or a key inside a dictionary.

Let me know what you think :)

### DIFF
--- a/src/invoice2data/output/to_csv.py
+++ b/src/invoice2data/output/to_csv.py
@@ -1,7 +1,6 @@
 import csv
 import sys
 
-
 def write_to_file(data, path, date_format="%Y-%m-%d"):
     """Export extracted fields to csv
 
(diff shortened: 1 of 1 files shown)


m3nu (approved):

m3nu:
Good to know. Merged your updated docs. Thanks for fixing!

Issue #302, opened 2020-09-19
options->replace not only replace literally values


rahxphoon:
Hi,
I recently use your smart library to do invoice extraction.
Since I cannot process multiple lines due to the limitation
If there is a way for me to replace like **L\)\r\n ->L\)** to reassemble multiple lines into one line  before processing?

Regards
Charles

PR #335, opened 2021-03-19
Replace changed to regex sub and added type hinting

RossK1:
Switched the replace functionality from .replace to re.sub. This should maintain backwards compatibility while allowing for more complex substitutions with regex patterns such as requested in issue #302.
Also added type hinting to the function definition.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -287,7 +287,9 @@ options and their defaults are:
 - `languages` (default = \[\]): Also passed to `dateparser` to parse
   names of months.
 - `replace` (default = `[]`): Additional search and replace before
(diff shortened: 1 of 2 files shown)


m3nu:
Fine to add regex replacements. Missing for this PR:

- Add documentation on the use of regex replacements
- If you want to add type annotations, they need to be added to CI as well to have an effect.

RossK1:
@m3nu, I expanded on the replace section in the tutorial to indicate that it now supports regex replaces :) The type hinting is more to be in line with PEP484 and others and make it slightly more readable for future work rather than meaning to be enforced.

m3nu (approved):

PR #336, opened 2021-03-21
Added log entries with table text and pattern

RossK1:
Added logger.debug entries to display the extracted table text and the regex pattern. Very useful when running a single template with numerous tables and want to check exactly what each was seeing.


m3nu (approved):

m3nu:
Thanks! Merged.

Issue #340, opened 2021-04-28
No template for Invoice/*******.pdf False


NayanMogra:
Hi, I created a template for my pdf, and when I try to run that it is again showing this no template found,
Only QualityHosting invoice is working other invoices are having the same issue though I am creating templates

Animesh-CSE:
Same Issue hare also

NayanMogra:
@Animesh-CSE  
I found the solution add this code in your py file 
import logging.config 
import logging 
logging.basicConfig(level=logging.DEBUG) logging.basicConfig()

And then it will display all the pdf data 
After that just compare that issuer name you wrote is same as that in terminal

Animesh-CSE:
@NayanMogra  Thank You
Now come another issues.Hare it is-

**Traceback (most recent call last):
  File "G:\project\project.py", line 6, in <module>
    result = extract_data('invoice/MyFlipkart.pdf',templates=templates)
  File "C:\Users\Hp\AppData\Local\Programs\Python\Python37-32\lib\site-packages\invoice2data\main.py", line 93, in extract_data
    return t.extract(optimized_str)
  File "C:\Users\Hp\AppData\Local\Programs\Python\Python37-32\lib\site-packages\invoice2data\extract\invoice_template.py", line 178, in extract
    output[k] = self.parse_date(res_find[0])
  File "C:\Users\Hp\AppData\Local\Programs\Python\Python37-32\lib\site-packages\invoice2data\extract\invoice_template.py", line 113, in parse_date
    value, date_formats=self.options['date_formats'], languages=self.options['languages']
  File "C:\Users\Hp\AppData\Local\Programs\Python\Python37-32\lib\site-packages\dateparser\conf.py", line 89, in wrapper
    return f(*args, **kwargs)
  File "C:\Users\Hp\AppData\Local\Programs\Python\Python37-32\lib\site-packages\dateparser\__init__.py", line 54, in parse
    data = parser.get_date_data(date_string, date_formats)
  File "C:\Users\Hp\AppData\Local\Programs\Python\Python37-32\lib\site-packages\dateparser\date.py", line 412, in get_date_data
    raise TypeError('Input type must be str')
TypeError: Input type must be str**

RossK1:
Can you share your template here, please?

NayanMogra:
@Animesh-CSE  yes please share your template

Animesh-CSE:
@RossK1 
@NayanMogra 

[...]

Issue #339, opened 2021-03-25
Debugging a template file is too difficult


reikred:
Alright, I'm going to vent a bit here: I have been an occasional user of i2d for a few months now, and I find that template development is really difficult.  Here are some of the difficulties I run into:

1. When there is an error, I just get a stack dump. No name of the template file used, no line number for the template file, no line number for the data file (if a runtime error rather than a parse error), no offending regex, nothing. More info really is needed. Debugging this stuff is a hard for me as debugging elisp :-).

2. In other words, I can't even tell which template file and variable or line or regex in the template file is the offending entity

3. The TUTORIAL.md is fairly incomprehensible and in my view is not written well (say, compared to a GNU manual or similar).  I do not gain any insight into how the program designer was thinking, which is usually the key to understanding how to use a   program properly. When I read this "issues" forum, I see a lot of allusions to the correspondence between template syntax and various internal data structures of the program, but my understanding is lacking. Maybe I should not need to understand, but because the error messages are so cryptic I am grasping for a way to understand syntax and semantics.

4.  The --debug flag just dumps the pdftotext (or equivalent) to stderr, it does not appear to provide any other useful debug output. 

I freely admit that I am not a good python hacker, but as an experienced linux user, decently versed in regexp, and an occasional scripter/programmer , I thought I would understand SOMETHING from the error messages, but generally I do not. For example, what on earth am I supposed to understand from the following error message?

% invoice2data --template-folder ./tpl 86435364.pdf
Traceback (most recent call last):
  File "/usr/local/bin/invoice2data", line 11, in <module>
    load_entry_point('invoice2data', 'console_scripts', 'invoice2data')()
  File "/home/reikred/src/invoice2data/src/invoice2data/main.py", line 211, in main
    res = extract_data(f.name, templates=templates, input_module=input_module)
  File "/home/reikred/src/invoice2data/src/invoice2data/main.py", line 94, in extract_data
    return t.extract(optimized_str)
  File "/home/reikred/src/invoice2data/src/invoice2data/extract/invoice_template.py", line 178, in extract
    res_find = re.findall(v, optimized_str)
  File "/usr/lib/python3.8/re.py", line 241, in findall
    return _compile(pattern, flags).findall(string)
  File "/usr/lib/python3.8/re.py", line 294, in _compile
    return _cache[type(pattern), pattern, flags]
TypeError: unhashable type: 'collections.OrderedDict'

[...]

Issue #149, opened 2018-07-19
ERROR:invoice2data.main:No template for invoice2.pdf


natashakhatib23:
I have installed invoice2data and pdftotext but i'm facing this error whenever i write "invoice2data invoice2.pdf" why?

duskybomb:
This because none of the existing templates match your pdf. You need to create your own template to make it work. You should go through [Tutorial for template creation](https://github.com/invoice-x/invoice2data/blob/master/TUTORIAL.rst)

natashakhatib23:
which means that invoice2data module isn't a general solution for data  extraction of the invoice.
Suppose i have an invoice from an organisation that i didn't create its template yet, it won't get me the data out of it .. right ?

natashakhatib23:
do you think machine learning algorithms would help in extracting data from an invoice with no template ?

m3nu:
Yes. Being done by the guys who got in touch in #148. I didn't test it, so I'm not sure about the accuracy. Since accounting processes generally demand 100% accuracy strict templates that fail when there is any uncertainty are possibly preferable for now.

In the medium term we strive to embed the invoice data directly and skip all the recognition. See [this other project](https://github.com/invoice-x/factur-x) we are working on for details.

darkar18:
same i am having same issue: using invoice2data module which i ran using : invoice2data --input-reader tesseract --template-folder \invoice2data\src\invoice2data\extract\templates aug.pdf

issuer: VILANGATTIL HARDWARES
keywords:
  - ERUVELI, CHOTTANIKKARA
  - 32AKUPJ3251Z4
fields:
  amount: (\d+)
  date:
    (\d{2})+\/+(\d{2})+\/+(\d{4})
  invoice_number: Invoice No.
 
    (\d{4})+\-+(\d{2})+\/+(.*[A-Z][0-9]+)
options:
  remove_whitespace: false
  date_formats:
    - '%d/%m/%Y'
lines:
    start: \#Item\snameHSN\/SACQuantityUnitPrice\/unitGSTAmount
    end: Total
    body: (?P<list>^\d{1,2})+\s+(?P<product>.*[A-Z])+\s+(?P<HSN>\d{4,8})+\s+(?P<quantity>\d+)
    
![Annotation 2021-09-14 211733](https://user-images.githubusercontent.com/65691792/133290813-854f2f21-c90c-4359-937b-56022cc71a06.png)

attaching: sample invoice for your reference

[...]

PR #355, opened 2021-09-17
Be more verbose about loaded templates

rmilecki:
```
People seem to be often confused by invoice2data CLI behaviour. When
something doesn't work the default output doesn't seem to be enough to
understand what's happening.

Try to improve that situation be printing info about templates that were
loaded and the used one. Use info level messages for that basic stuff so
it's clear what is happening even without the --debug switch.

```

### DIFF
--- a/src/invoice2data/extract/loader.py
+++ b/src/invoice2data/extract/loader.py
@@ -113,4 +113,7 @@ def read_templates(folder=None):
                     tpl["exclude_keywords"] = [tpl["exclude_keywords"]]
 
                 output.append(InvoiceTemplate(tpl))
+
+    logger.info("Loaded %d templates from %s", len(output), folder)
+
     return output
--- a/src/invoice2data/main.py
+++ b/src/invoice2data/main.py
@@ -85,11 +85,11 @@ def extract_data(invoicefile, templates=None, input_module=pdftotext):
     logger.debug(extracted_str)
     logger.debug("END pdftotext result =============================")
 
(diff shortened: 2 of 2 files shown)


rmilecki:
I thought we need to output more info by default after seeing reports like:

- https://github.com/invoice-x/invoice2data/issues/340
- https://github.com/invoice-x/invoice2data/issues/339
- https://github.com/invoice-x/invoice2data/issues/149#issuecomment-919281862

Examples:
```
INFO:invoice2data.extract.loader:Loaded 141 templates from /home/rmilecki/projects/invoice2data/src/invoice2data/extract/templates
ERROR:invoice2data.main:No template for document.pdf
```
```
INFO:invoice2data.extract.loader:Loaded 37 templates from /home/rmilecki/templates
INFO:invoice2data.extract.loader:Loaded 141 templates from /home/rmilecki/projects/invoice2data/src/invoice2data/extract/templates
INFO:invoice2data.main:Using allnet.yml template
INFO:invoice2data.main:{'issuer': 'ALLNET GmbH Computersysteme', 'amount': 93.83, 'invoice_number': '1234567', 'date': datetime.datetime(2020, 1, 21, 0, 0), 'kundennr': '12345', 'currency': 'EUR', 'desc': 'Invoice from ALLNET GmbH Computersysteme'}
```
Let me know if you think adding such messages it a good idea.

m3nu (approved):

PR #356, opened 2021-09-17
Simplify logging output for non-debug run

rmilecki:
```
Logged messages should be simple & easy to read when not running in a
debug mode. People running it for non-development purposes don't need to
know what internal classes were used.

Use the simplest logging format unless --debug is used.
```

### DIFF
--- a/src/invoice2data/main.py
+++ b/src/invoice2data/main.py
@@ -192,7 +192,7 @@ def main(args=None):
     if args.debug:
         logging.basicConfig(level=logging.DEBUG)
     else:
-        logging.basicConfig(level=logging.INFO)
+        logging.basicConfig(level=logging.INFO, format="%(message)s")
 
     input_module = input_mapping[args.input_reader]
     output_module = output_mapping[args.output_format]


rmilecki:
I think that for standard use printing any debugging info just makes it just harder to follow logged messages.

Let me know if you think this change is a good idea.

## Before
```
INFO:invoice2data.extract.loader:Loaded 37 templates from /home/rmilecki/templates
INFO:invoice2data.extract.loader:Loaded 141 templates from /home/rmilecki/projects/invoice2data/src/invoice2data/extract/templates
INFO:invoice2data.main:Using allnet.yml template
INFO:invoice2data.main:{'issuer': 'ALLNET GmbH Computersysteme', 'amount': 93.83, 'invoice_number': '1234567', 'date': datetime.datetime(2020, 1, 21, 0, 0), 'kundennr': '12345', 'currency': 'EUR', 'desc': 'Invoice from ALLNET GmbH Computersysteme'}
```

## After
```
Loaded 37 templates from /home/rmilecki/praca/healthy/ocr/tpl
Loaded 141 templates from /home/rmilecki/projects/invoice2data/src/invoice2data/extract/templates
Using allnet.yml template
{'issuer': 'ALLNET GmbH Computersysteme', 'amount': 93.83, 'invoice_number': '1234567', 'date': datetime.datetime(2020, 1, 21, 0, 0), 'kundennr': '12345', 'currency': 'EUR', 'desc': 'Invoice from ALLNET GmbH Computersysteme'}
```

m3nu (approved):

PR #357, opened 2021-09-19
TUTORIAL: Move up "Parser regex" header & first paragraph

rmilecki:
```
This way paragraphs describing regexes get into the "Parser regex"
section which is where they belong.

```


rmilecki:
## Before
before

## After
after

m3nu (approved):

Issue #271, opened 2020-01-15
Invoice with multiple pages


schmidtkropp:
Hello everybody,
I would like to use invoice2data for a small workaround with our accounting software.

I have a lot of invoices that span several pages. Unfortunately, the text is only used on the first page.
The invoices are scanned and then evaluated via tessercast:

invoice2data --input-reader tesseract /srv/kicArchiv_2020-01-142.pdf --template-folder / srv / templates --debug

Unfortunately, it often happens that the invoice amount is then on the second or third page.

m3nu:
It should work for multiple pages. The German Qualityhosting template was developed for multiple pages IIRC.

It's possible that Tesseract only does the first page. In that case it's just a setting and could be changed with small adjustments.

schmidtkropp:
Hello m3nu,

many thanks for the answer.
Can you describe the adjustments for Tessercast in more detail?

Twister42:
Unfortunately invoice2data convert PDF to images using the PNG format which do not handle multi pages. If you want to do multi page on PDF you have to change to code by yourself and use TIFF format : in tesseract.py line 37 : png -> tiff

It works fine for me. Looking at the comments in the file it seems they used tiff in the past but switched to png. don't know why...

bosd:
If i remember, previously the ghostscript engine was used for conversion.
Due to some security issues, the conversion was disabled, users had to bypass these rules.

I think it is fixed now with the current png implementation.
Now internally all the pages are added in one big png.
From experience png delivers as good as result then tiff files.
Hence, smaller file size, no security issues, works out of the box.

Issue #169, opened 2018-10-02
Pass clean image to Tesseract to avoid conversion warnings


netson:
Hi, first of all, thanks for a great module! :-) By spending a couple of hours creating templates for my documents, I will be saving hours each month! :+1: 

However, some of my documents are scanned, so I am using tesseract to extract the text from it. This works fine and as expected, but it does generate some errors on stderr:

```
convert-im6.q16: profile 'icc': 'RGB ': RGB color space not permitted on grayscale PNG `/tmp/magick-47439a8uUu8VvEXM' @ warning/png.c/MagickPNGWarningHandler/1654.
Warning. Invalid resolution 0 dpi. Using 70 instead.
Estimating resolution as 492
```

I'm running the invoice2data module from a custom python script which does everything I need it to in 1 go, so I'd like to see these error messages go away.

The first one seems to indicate that setting an RGB profile for grayscale images is not allowed; I haven't been able to figure out what causes the other messages yet (could be due to my lack of imagemagick skills).

My system:
```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
Python 2.7.15rc1
Invoice2data 0.2.101.dist
```

I could simply suppress the stderr messages (as mentioned, the module works as expected), but that seems wrong and there is no clean way to do this from within python without modifying the module directly.

Any idea what the issue could be? And can I fix this with settings or is it a bug in the module?
Any help is much appreciated!

m3nu:
Those messages come from Tesseract. If there is a switch to turn them off, we can add them to `invoice2data`.

m3nu:
Does this issue persist when using the tensorflow4 output plugin?

bosd:
This issue is because how this module pre-processes the image.
The document is converted to the tiff format along with some other parameters.
However, imagemagick is not very good at passing the resolution data found in the exif of the input file to the output file in a different format.
(also some output formats don't allow exif data.)

The conversion code is in:
https://github.com/invoice-x/invoice2data/blob/36863386692a2393d561cdaa1e3a5e4938c34935/src/invoice2data/input/tesseract.py#L27-L38

From my testing, I got better ocr results by passing my png files directly to tesseract.
In my opinion, The flattening of the image in this code does more harm than good.
Tesseract uses leptonic to read/pre-process the image.

yet. the conversion code is required for other file formats as tesseract can't handle pdf files as an input directly.

bosd:
[...]

PR #373, opened 2022-03-27
Multipage PDF support, Pass Image files cleanly

bosd:
Fixes: 
Multi pdf support https://github.com/invoice-x/invoice2data/issues/271
pass clean image https://github.com/invoice-x/invoice2data/issues/169

all pages from a PDF file get converted to a png image before feeding it into tesseract.
Because the convert function of Imagemagick loses the dpi information, it is set to 300 dpi.

### DIFF
--- a/src/invoice2data/input/tesseract.py
+++ b/src/invoice2data/input/tesseract.py
@@ -7,12 +7,12 @@ def to_text(path):
     Parameters
     ----------
     path : str
-        path of electronic invoice in JPG or PNG format
+        path of electronic invoice in JPG, PNG or PDF format
 
     Returns
     -------
     extracted_str : str
-        returns extracted text from image in JPG or PNG format
+        returns extracted text from image or PDF file
 
     """
     import subprocess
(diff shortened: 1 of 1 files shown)


bosd:
@RossK1 Would you be so kind to review?

bosd:
@m3nu Can you trigger tests?

bosd:
force pushed to trigger tests

bosd:
@m3nu Can you see the button?

bosd:
@rmilecki Can you review this one? :pray:

bosd:
Removed the filename change from this code.
As I think the code in Pr #384 is better.
So this one is now only for the fixes and improvement of OCR.
Hope this one can be merged quickly (into a new release).
@m3nu 🙏

src/invoice2data/input/tesseract.py
```diff
@@ -1,3 +1,4 @@
+
```
rmilecki, line 1: Nitpicking: I don't think we need that extra empty line?

rmilecki:
Looks fine I guess.

rmilecki (approved):

PR #379, opened 2022-06-23
Unittests use try except for pdfminer_wrapper

bosd:
Use similar method as for pdftotext unittests.

### DIFF
--- a/tests/test_lib.py
+++ b/tests/test_lib.py
@@ -72,7 +72,7 @@ class TestCLI(unittest.TestCase):
             except ImportError:
                 # print("pdfminer module not installed!")
                 self.assertTrue(False, "pdfminer is not installed")
-                self.assertTrue(type(res) is dict, "return is not a dict")
+                self.assertTrue(type(res) is str, "return is not a string")
 
     def test_tesseract_for_return(self):
         png_files = get_sample_files('.png')


bosd:
force pushing to trigger tests

m3nu:
All approved. Hopefully we can count you as contributor soon.

bosd:
@rmilecki Can you please review?

tests/test_lib.py
```diff
@@ -66,7 +66,13 @@ def test_output_csv(self):
...
         pdf_files = get_sample_files('.pdf')
         for file in pdf_files:
-            extract_data(file, None, pdfminer_wrapper)
+            try:
+                res = extract_data(file, None, pdfminer_wrapper)
+                print(res)  # Check why logger.info is not working, for the time being using print
+            except ImportError:
+                # print("pdfminer module not installed!")
+                self.assertTrue(False, "pdfminer is not installed")
+                self.assertTrue(type(res) is dict, "return is not a dict")
```
rmilecki, line 75: I think that `return is not a dict` test should be out of the `except` block

rmilecki:
Looks good except for a bit messy commits. We have some (unneeded?) merge commit and commit adding `extract_data()` call.

If all those changes can be squashed into single commit during merge - that's great. If not maybe we could have that branch rebased and commits squashed manually?

bosd:
> Looks good except for a bit messy commits. We have some (unneeded?) merge commit and commit adding `extract_data()` call.

Yes, I was fighting with the git cli.
Seems ok, now

rmilecki (approved):

bosd:
Moved the test to the try block. As there is no result to test if the pdfminer module is not installed.


Yet, don't quite understand why it is failling. As using similar method as the pdftotext test.

Will have to dig deeper into this one.
Any ideas?

bosd:
Hmm.. tests are still failling.
Where is defined that pdfminer should be installed for the test suite?

bosd:
✨ All ✔️ now!!
@m3nu Can you merge?

bosd:
@m3nu  Do you know how to fix this one?
It seems stuck on waiting for the result of the python 3.6 test.
Which propably will never come in, as we removed it from the github workflow.

m3nu:
Python 3.6 needed to be removed as required status check in GH Actions settings.

PR #382, opened 2022-06-24
test_cli Fix Assertion error due to path mismatch

bosd:
running test_cli locally results in an assertion error, because it is testing against the wrong folder.

steps to reproduce:
1. go to by `$ cd invoice2data/tests`
2. `$ python test_cli.py`

results in

```
======================================================================
FAIL: test_copy (__main__.TestCLI)
----------------------------------------------------------------------
  File "test_cli.py", line 136, in test_copy
    self.assertEqual(i, len(get_sample_files('.json')))
AssertionError: 0 != 6

----------------------------------------------------------------------
```
This fix let it test against the created folder. By using the same variable.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -126,7 +126,7 @@ class TestCLI(unittest.TestCase):
         main(args)
         i = 0
         for path, subdirs, files in os.walk(
-            pkg_resources.resource_filename(__name__, 'copy_test/pdf')
+            pkg_resources.resource_filename(__name__, directory)
         ):
             for file in files:
                 if file.endswith(".pdf"):


bosd:
force pushing to trigger tests

rmilecki:
@bosd: does this problem (AssertionError) occur if you run tests using `python -m tests.test_cli` ?

bosd:
> does this problem (AssertionError) occur if you run tests using `python -m tests.test_cli` ?

Just tested both cases:
1. Before this commit. Running `python -m tests.test_cli` returns OK.
2. After this commit `python -m tests.test_cli` returns FAILED :exclamation: .

So we can conclude that:
1. Better not merge this commit.
2. Call unittests with: 
`python -m tests.test_cli`
`python -m unittest tests/test_cli.py`
`python -m unittest discover`

 Better improve documentation how to properly run the tests.

As a side note. Would expect that `python -m unittest invoice2data` would work. But it actually won't call any test.

Issue #383, opened 2022-06-28
Json output in wrong order


bosd:
The written content to the json file is in the wrong order.
Environment: local system. Might occur on CI unit tests as well 

Steps to reproduce:
invoice2data ./SammyMaystoneLinesTest.pdf --input-reader=pdftotext --template-folder=./Templates --output-format=json --output-name=samy-before.json --debug

Debug output:

```
{'issuer': 'Sammy Maystone - For Lines testing', 'invoice_number': 'invoice_number_1', 'date': datetime.datetime(2022, 1, 1, 0, 0), 'line_items': [{'item': 'A', 'desc': 'Parts: 1 x cap_a'}, {'item': 'B', 'desc': 'Parts: 2 x shop supplies'}], 'currency': 'USD', 'desc': 'Invoice from Sammy Maystone - For Lines testing'}
```

Actually written to the json file:

```
[
    {
        "currency": "USD",
        "date": "2022-01-01",
        "desc": "Invoice from Sammy Maystone - For Lines testing",
        "invoice_number": "invoice_number_1",
        "issuer": "Sammy Maystone - For Lines testing",
        "line_items": [
            {
                "desc": "Parts: 1 x cap_a",
                "item": "A"
            },
            {
                "desc": "Parts: 2 x shop supplies",
                "item": "B"
            }
        ]
    }
]
```

This causes the unittests to fail when it is compared against
https://github.com/invoice-x/invoice2data/blob/master/tests/compare/SammyMaystoneLinesTest.json

Encountered this while testing for: https://github.com/invoice-x/invoice2data/issues/378

Output to CSV, and Library are ok.
Untested: output to XML.

bosd:
@rmilecki Any thoughts on this?

bosd:
@m3nu Your thoughts on this?

rmilecki:
That comes from `sort_keys=True` in the `src/invoice2data/output/to_json.py`. I think we have two options:

1. Drop `sort_keys=True`
2. Update `SammyMaystoneLinesTest.json` (sort its keys)

I think that `sort_keys=False` would be more natural but it's also a behaviour change. On the other hand it's a very minimal change that noone should complain about.

@m3nu: thoughts?

bosd:
IMO we should drop the sort_keys

PR #394, opened 2022-09-15
[FIX] Unittests: disable sort_keys in JSON output

bosd:
As discussed in #383

Unittests are failling because the output is in the wrong order.
From an usability pov. We want to output keys just as they were parsed from the invoice.
Eg in lines section:
Section_header
Productline 1
Productline 2

instead of (alphabetical order)
Productline 1
Productline 2
Section_header


rmilecki:
Looks good, thanks!

Issue #374, opened 2022-04-08
Output filename and Move function not working [AttributeError]


bosd:
Running invoice2data with the `--move` option on  python3 throws the following error.

```
Traceback (most recent call last):
  File "/usr/local/bin/invoice2data", line 10, in <module>
    sys.exit(main())
  File "/home/emiel/.local/lib/python3.7/site-packages/invoice2data/main.py", line 227, in main
    date=res["date"].strftime("%Y-%m-%d"),
AttributeError: 'list' object has no attribute 'strftime'
```

I have noticed that in main.py datetime is not imported.
adding the import did not solve the prolem.

bosd:
Also noticed that by default the output extension gets forced to pdf. Even when processing scanned image files.

PR #384, opened 2022-06-30
Fix issue #316: Add support for all template variables in filename.

sandervd:
First of all, thank you for this project!

This PR adds support for any template variable to be used in the filename as requested in #316.

PR #403, opened 2022-09-24
support more formatting with --filename-format option

rmilecki:
```
Before this change --filename-format accepted string with up to three
variables only ("date", "invoice_number" and "desc"). It also couldn't
deal with "date" being a list of dates.

This change allows using any parsed field. It supports date in any
field. It also deals with list of dates by picking the first one.

Example usage:
--filename-format "{date} {issuer} {invoice_number} {currency}.pdf"
```

### DIFF
--- a/src/invoice2data/main.py
+++ b/src/invoice2data/main.py
@@ -2,6 +2,8 @@
 # -*- coding: utf-8 -*-
 
 import argparse
+import copy
+import datetime
 import shutil
 import os
 from os.path import join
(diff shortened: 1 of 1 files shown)


rmilecki:
I'd like to propose this change as an alternative solution to both:
https://github.com/invoice-x/invoice2data/pull/375
https://github.com/invoice-x/invoice2data/pull/384

This also fixes https://github.com/invoice-x/invoice2data/issues/374

bosd (approved):
tested locally, LGTM :+1: 
(did not test is with a dict of dates)

@rmilecki 
This rewrites the filename of the invoicefile when it is moved or copied. (thus PDF file).
It is not a function I'm likely going to use. But for consistency does it make sense to also have the same options for the csv/json... output files?
If so, I think it should be a sepearte pr.

PR #405, opened 2022-09-24
print which file pdfminer is testing

bosd:


rmilecki:
@bosd: could you in general give others some time to review changes, please? Wait 1-2 days before merging pull requests?

This change looks OK of course, it's just a general request.

bosd:
Sure will do, for PR's that are worthwhile.
For minor fixes (like this), updating comments and small doc changes I prefer to keep the workflow. Better to be sorry then delay. 😉

PR #406, opened 2022-09-25
tests: print file for which content verification failed

rmilecki:
```
tests: print file for which content verification failed

This helps tracking & fixing regressions during development.

Example output:
======================================================================
FAIL: test_content_json (__main__.TestCLI)
----------------------------------------------------------------------
    File "/home/rmilecki/projects/invoice2data/tests/test_cli.py", line 78, in test_content_json
    self.assertTrue(False, 'Failed to verify parsing result for ' + jfile)
AssertionError: False is not true : Failed to verify parsing result for /home/foo/invoice2data/tests/compare/AmazonWebServices.json
```


bosd (approved):
LGTM :+1:

Issue #393, opened 2022-09-12
image to data


aditya11ad:
hi,

is there any way to pass "invoice **image**".
I mean can we apply tesseract ocr on the image first and then invoice2data.

thanks in advance.

bosd:
Ran into the same problem.
Some invoices have crucial data encaptured in images.

This library had either "text extraction" or "ocr".

What you are looking for is propably an HOCR function.
Which applies OCR to the images and extracts the texts from the PDF.

This library supports pdfminer.
Recently pdfminer got hocr support.

pdfminer/pdfminer.six#651

Didn't have time yet to play with that functionality, and how it can be used in this library.
Maybe we need to update documentation how to use it.

bosd:
Sorry, my previous statement was wrong.
Best method for now, is to pass the whole invoice to the tesseract input method.

bosd:
@aditya11ad 
I've made a pr, in which ocrmypdf is used as a pre-processor for invoice2data.
Can you review that one? Check if it covers your use case?

bosd:
Implemented in #409

PR #409, opened 2022-09-25
[WIP] Saeco Example text & OCR to data

bosd:
This is an example of a PDF with some vital data encapsulated in a image.
Like mentioned in #393

The current strategy to parse this kind of file is to parse the whole thing with tesseract.
However this is not the preffered method. Because even the original text elements get parsed with tesseract.
With the acompanied errors, it misread a `0` (zero digit) for a `O` (letter).

Adapted the template to workaround this.
It get's the job done.

But it would be better if we had an solution like:
- Pass only the images to OCR
- Parse the texts from the PDF directly.

Additional context.
While using invoice2data as a library. In it currents form it is parsing the pdf with the pdftotext parser.
If the parser files, then it calls invoice2data again with the tesseract as a parser.
So I've changed the template to delebiratly fail when using the pdftotext parser.


bosd:
The solution is here :smile: :sparkles: 

Added [ocrmypdf](https://github.com/ocrmypdf/OCRmyPDF) as an input module. (it is actually more of a pre-processor)
Now we are able to parse this kind of mixed-type invoice.

ocrmypdf has a function to redo-ocr. This is the function we are calling by default for invoice2data. It will leave the exsisting texts unchanged. Images which not have been ocr'ed yet will be processed. 

Still, there is the problem, that the pdf needs to be ocr'ed before keywords can be matched.
To solve this issue, ocrmypdf is now a fallback module.

It works like this:
If the default pdftotext input parser failes to match a template. It will check if ocrmypdf is installed.
If it is installed, ocrmypdf is called with the redo-ocr parameters. The result of ocrmypdf will be sent trough pdftotext again and tries to find an template again. The template is found, and extraction of the pdf will start.
See the Saeco example.

ocrmypdf also has a lot of functionality to bring to the table.

- Generating a [PDF/A](https://en.wikipedia.org/?title=PDF/A) file.
- Optimizes the PDF images, to reduce file size
- deskews and/or cleans the image before performing tesseract OCR (Very handy for scanned tickets and receipts)
it uses [unpaper](https://github.com/Flameeyes/unpaper/blob/main/doc/basic-concepts.md) for that

Personally I was using unpaper in my pipeline to process scanned receipts before sending them to my invoice2data server. Having the possibility to let invoice2data do it makes it much easier.

You can fully control the parameters of ocrmypdf and it's underlying modules (unpaper, tesseract) by passing your parameters in the `input_reader_config` dict.

bosd:
Test coverage is dropping because of this one. So I will add some more tests.

bosd:
Coverage is 80% + This should do.

bosd:
@rmilecki Can you review this pr? :pray:

bosd:
This PR introduces some important functionality. It closes the gaps of mixed type pdf's which previsously could not been parsed.
I'm keen for getting it into production.  (Which pulls invoice2data form pypi)
Can either @rmilecki or @m3nu approve so we can merge this.

PR #410, opened 2022-09-26
Check if any text was extracted before trying to parse it

rmilecki:
```
This prevents further parsing errors & provides some meaningful info to
user on the problem.
```


bosd (approved):
Quick code review!
LGTM :+1: 
Got to to make sure that every input module returns a empty string.
And not a `Nonetype`
That would result in a
`AttributeError: 'NoneType' object has no attribute 'strip'`

rmilecki:
@bosd now you pointed that out, I think it actually makes sense to let input modules return other types too (`None`? `False`?). Maybe I could make that check safer to allow that and open way for input modules improvements.

Issue #151, opened 2018-07-23
Use with plain text input


rmnppt:
Hi Guys,

Not sure this is the right place for this...

I would like to try out your library - looks great! Only issue is that I have a plain text (sometimes html) dataset. Is there an input workflow for data other than .pdf?

Thanks in advance for your advice!

m3nu:
Plain text and HTML should clearly work. It would just need a simple input module to open and parse the file. For HTML it could remove the tags and keep just the text.

m3nu:
For HTML you will be better off using something like CSS or Xpath to match your data. Since you have the HTML structure to use. This project is for unstructured input.

Of course HTML could be unstructured. E.g. a plain text file or a very badly formatted HTML. In that case we could add a plain text input module that just passes the input on. But I don't see much demand for this.

rmilecki:
Implemented in the f59d6838db62dd6c322f99cf274e7f5687adb00d ("input: add support for plain text")

PR #411, opened 2022-09-26
Support parsing plain text invoices (like invoice.txt)

rmilecki:
```
tests: add test for parsing Orlen invoices

This is the first test invoice in text format.
```
```
tests: add support for comparing text input files with expected output

Now that invoice2data supports plain text invoices (with the "text"
input module) we should also include some tests for it to make sure this
feature is properly maintained.
```
```
Add very basic (pdftotext vs. text) auto-detection of input file format

We can't always detect the best input module as there are e.g. multiple
ones for handling OCR. We can however decide between "text" and
"pdftotext" as default.

This makes using invoice2data a bit easier for parsing plain text
invoices.
```
```
input: add support for plain text

This extends possible invoice2data usage for dealing with plain text
documents.

It may be also useful for:
1. Debugging - when reporter provides parsed PDF content only
2. Testing - it may be easier to write more test for plain text files

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ won't parse tables in PDF correctly.
 Basic usage. Process PDF files and write result to CSV.
 
 - `invoice2data invoice.pdf`
+- `invoice2data invoice.txt`
 - `invoice2data *.pdf`
 
 Choose any of the following input readers:
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A command line tool and Python library to support your accounting
 process.
 
 1. extracts text from PDF files using different techniques, like
-   `pdftotext`, `pdfminer` or OCR -- `tesseract`, `tesseract4` or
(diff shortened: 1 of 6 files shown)


rmilecki:
This implements feature requested in the https://github.com/invoice-x/invoice2data/issues/151

I took care of updating documentation and adding test.

This may be also useufl if we prefer to store some virtual (custom) invoices internally (for testing purposes). It may be easier to store them in text format.

rmilecki:
@bosd: does it look OK now?

@m3nu: would you like me to wait some specific amount of time before I merge such new features? To give you time for review?

bosd:
By looking at the code it seems OK.
Tests are passing so it is a :+1: for me

rmilecki:
@m3nu: I wanted 3 full days since my last update to this pull request and merged it today. I hope it's OK and you don't mind me doing that.

PR #412, opened 2022-09-26
Fixup mention tesseract4 input module

bosd:


bosd:
Do not merge yet.
I think the tesseract modules and tesseract4 needs a rewrite.

Current implementation is confusing. Because of the missing legacy argument.
The tesseract input module already calls tesseract4.

What the current tesseract4 module does is calling tesseract with different arguments. 
And applying different pre processing steps before feeding the file to tesseract.

Those pre-processing steps are currently broken due to security restrictions in the postscript and imagemagick module.

bosd:
Closing this a rewrite of tesseract modules is in different pr.

Issue #181, opened 2018-12-03
Language as a input parameter


suryacaprice:
I can see the default language is french. But can you please provide option to enter language as input so that english or other languages can be switched easily .

m3nu:
Where is French the default language?

m3nu:
I see. Yeah. I did this for a French client.

If you want an option, just open a pull request and I'll review it.

suryacaprice:
Sure will do

knutole:
Any progress on this? Very useful to be able to pass language to tesseract for example.

bosd:
This is a tricky problem to tackle. As one have to know and input the language before the rpcessing takes place.
It quickly leeds to a complicated pipline
[Preprocess image (ocr) to determine language](https://github.com/JAIJANYANI/Language-Detection-in-Image) --> match detected words against dictionary to guess language --> process the file again with the guessed (detected)  language.



I would like to propose a more simple solution.
Set the default language to English. (as it is internationally much more common then french.)
Poll which additional tesseract ocr languages are installed by the user.

Process the document with the default language + all the installed languages of the user. 
Tesseract will then use the language which will best fit the image.



Proposed solution will work for tesseract, as it is used in the [OCR mypdf](https://github.com/ocrmypdf/OCRmyPDF/tree/master/src/ocrmypdf) project.

https://github.com/ocrmypdf/OCRmyPDF/blob/3dfde479e2d0d6dac4c6f55093038068e475b82c/src/ocrmypdf/builtin_plugins/tesseract_ocr.py#L154-L155

bosd:
Fixed by the above 2 PR's

PR #413, opened 2022-09-26
set gvision default lang to english

bosd:
as discussed in: #181


bosd:
> I guess English makes more sense but it's probably still not perfect to hardcode it.

True, I think it is good to open up a discussion (sometime) about the configurability / paramaterization of input modules, plugins. It's quite a big topic, better to discussed elsewhere. I waited with opening it, because I want to show it with an example. With the Camleot plugin, the json configurations is passed trough the yaml template.. More on that later.

On this PR:
This was just a quick fix / purposal.
I did not try to use gvision with this library, As my of the open PR's and issues, it seems like current implementation is broken.
So, it might need some attention.

rmilecki (approved):
Well, it looks good probablt, as English makes more sense I guess.

@m3nu: any opinion on this?

PR #414, opened 2022-09-27
tests: prepare for custom invoices and templates

rmilecki:
```
There is limited amount of real invoices included in the project and not
everything can be tested using them. To extend tests coverage add place
for storing custom invoices & templates. That way we can write more
complex templates that use more complex syntax.
```


rmilecki:
This depends on https://github.com/invoice-x/invoice2data/pull/411

bosd (approved):
LGTM :+1:

PR #415, opened 2022-09-27
tests: add custom test for parsing lines

rmilecki:


rmilecki:
@bosd: I'll use this branch & pull request to collect & verify our expectations for the extended `lines` syntax

I'll keep working on this during next days, please give me some extra time for that.

bosd:
Looks good! 

Not sure if my objective on the line parser is clear.
The challenge I was facing is mixing matching line rules, and maintaining the order as they appear in the input file.

eg.  with this example that would be:
```
No Name Quantity Total Price
 1 Foo  100      1000
 2 Bar         90
 3 Qux  5        50

```

In above example on line 2, there is no digit given for the quantity.
So the first regex wil fail:
`       line: ^\s*(?P<pos>[\d]+)\s+(?P<name>.+?)\s+(?P<qty>\d+)\s+(?P<total>\d+)`

But this second line rule will match..
`        line: ^\s*(?P<pos>[\d]+)\s+(?P<name>.+?)\s+(?P<total>\d+)`

Expected output

```
[
    {
        "issuer": "Lines Tester",
        "date": "2020-10-31",
        "invoice_number": "123",
        "amount": 99.0,
        "lines": [
            {
                "pos": 1,
                "name": "Foo",
                "qty": 100,
                "total": 1000
            },
            {
                "pos": 2,
                "name": "Bar",
                "total": 90
            },
            {
                "pos": 3,
                "name": "Qux",
                "qty": 5,
                "total": 50
            }
        ],
        "currency": "EUR",
        "desc": "Invoice from Lines Tester"
    }
]

```


Dunno if that was the purpose of this PR. But I think the example with a totally different second matching rule is more clear.. 
Like in 407

rmilecki:
I think I can close this one as most interesting tests for `lines` are already present. They were added by:

- https://github.com/invoice-x/invoice2data/pull/423
- https://github.com/invoice-x/invoice2data/pull/427

Please note that with https://github.com/invoice-x/invoice2data/pull/417 we can now have different lines supported by different regexes.

So for your case @bosd you can just use:
```
    line:
      - ^\s*(?P<pos>[\d]+)\s+(?P<name>.+?)\s+(?P<qty>\d+)\s+(?P<total>\d+)
      - ^\s*(?P<pos>[\d]+)\s+(?P<name>.+?)\s+(?P<total>\d+)
```

PR #416, opened 2022-10-01
Support "priority" and add 2 templates for InsERT-software-generated invoices

rmilecki:
```
templates: pl: add templates for InsERT's software issued invoices

InsERT is one of the most popular Polish accounting software company.
They have two very common softwares:
1. Subiekt nexo
2. Subiekt GT

Add 2 templates to parse invoices generated by above softwares. Those
templates are software-specific so they have priority set to 3.

This allows parsing a lot invoices issued in Poland.
```
```
Regex: support more grouping functions

In case of multiple matches we only had an option to return a sum of
numeric values. Add more functions.
```
```
Add "priority" support for templates

In case of multiple templates matching given invoice - choose the one
with the highest "priority" value. To provide proper support for
prioritizing AND existing templates (backward compatibility) the default
value 5 is assumed in case "priority" property is missing.

This feature can be used for writing more generic as well as more
specific templates. So far all templates were assumed to be
company-specific. With this change we can have:
1. Invoice-generating software specific templates
2. In-company varying templates

This feature may be very useful for:
1. Countries with just few very popular accounting software applications
2. Big companies with multiple departments adding some invoice details
```

### DIFF
--- a/src/invoice2data/extract/parsers/regex.py
+++ b/src/invoice2data/extract/parsers/regex.py
@@ -49,6 +49,14 @@ def parse(template, field, settings, content, legacy=False):
     if "group" in settings:
         if settings["group"] == "sum":
             result = sum(result)
+        elif settings["group"] == "min":
+            result = min(result)
+        elif settings["group"] == "max":
+            result = max(result)
+        elif settings["group"] == "first":
+            result = result[0]
+        elif settings["group"] == "last":
+            result = result[-1]
         else:
             logger.warning("Unsupported grouping method: " + settings["group"])
             return None
--- a/src/invoice2data/extract/templates/pl/pl.insert.subiekt-gt.yml
+++ b/src/invoice2data/extract/templates/pl/pl.insert.subiekt-gt.yml
@@ -1,0 +1,43 @@
+# SPDX-License-Identifier: MIT
+keywords:
+  - 'Miejsce wystawienia:'
+  - 'Data wystawienia:'
+  - 'Sprzedawca:'
+  - 'Nabywca:'
+  - 'według stawki VAT'
+  - 'Razem do zapłaty:'
+  - 'Wystawił\(a\)'
+  - 'Odebrał\(a\)'
+  - 'Podpis osoby upoważnionej'
+fields:
+  issuer:
+    parser: regex
+    regex: Sprzedawca:.*\n(.*?)\s{3,}
+  vatin:
+    parser: regex
+    regex: NIP:\s+(\d{10})
+    type: int
+    group: first
+  date:
+    parser: regex
+    regex:
+      - Data wystawienia:\n.*(\d{2}\.\d{2}\.\d{4})
+      - Data wystawienia:\n.*(\d{4}-\d{2}-\d{2})
+    type: date
+  invoice_number:
+    parser: regex
+    regex: Faktura VAT\s+(.*?)\s+oryginał
+  amount:
+    parser: regex
+    regex: Razem do zapłaty:\s+([\d\s]+,[\d][\d])
+    type: float
+  nrb:
+    parser: regex
+    regex: PLN:\s+([0-9]{2}(?:\s?[0-9]{4}){6})
+options:
+  currency: PLN
+  date_formats:
+    - '%d.%m.%Y'
+    - '%Y-%m-%d'
+  decimal_separator: ','
+priority: 3
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -301,6 +301,21 @@ options and their defaults are:
   different fields, you can supply a list here. The extraction will
   fail if not all fields are matched.
 
+### Priority
+
+In case of multiple templates matching single invoice the one with the
+highest priority will be used. Default `priority` value (assigned if
+missing) is 5.
+
+This property needs to be specified only when designing some generic or
+very specific templates.
+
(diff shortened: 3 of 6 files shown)


bosd:
Thanks for this very interesting PR!

Will need some time to test it.
I like the idea of creating generic templates, even selecting on the accounting software or library that generated it.
At some point it might be interesting to look / filter on the meta data passed in the invoicefile. 

Just a thought..
Will the use of prioritys have a big impact on precessing time?

Either way, being able to parse multiple invoices and minimize the writing write specific templates is a big plus.

rmilecki:
I'm happy to get a positive respose to this. Take your time to review / test this.

> Will the use of prioritys have a big impact on precessing time?

I think at loading templates takes the most time. This part doesn't get changed with this pull request.

With this change `matches_input()` gets called for every loaded template. Previously it was called until the first match. `matches_input()` is a simple function that just uses `re.search()`. So this pull request shouldn't affect perfroamnce much.

---

Since you mentioned perfromance I think we should check why loading templates takes so much time. Is that file access or parsing YAML? Maybe we could optimize / cache?

(Just a topic for another discussion)

src/invoice2data/extract/parsers/regex.py
```diff
@@ -46,6 +46,14 @@ def parse(template, field, settings, content, legacy=False):
     if "group" in settings:
         if settings["group"] == "sum":
             result = sum(result)
+        elif settings["group"] == "min":
```
bosd, line 49: Very interesting what your doing there.
Would this also work for date fields??

Currently when, a reges matches multiple date fields, and dict error occurs.
Would be interesting to prevent these dict errors.
e.g. Being able to return the oldest date as invoice date. Latest date as the due date.

src/invoice2data/extract/templates/pl/pl.insert.subiekt-gt.yml
```diff
@@ -0,0 +1,43 @@
...
+  - 'według stawki VAT'
+  - 'Razem do zapłaty:'
+  - 'Wystawił\(a\)'
+  - 'Odebrał\(a\)'
+  - 'Podpis osoby upoważnionej'
+fields:
+  issuer:
+    parser: regex
+    regex: Sprzedawca:.*\n(.*?)\s{3,}
+  vatin:
```
bosd, line 16: I know there is no settlement yet on standard field names.
But is it ok, for you to rename this to `vat`?

src/invoice2data/extract/templates/pl/pl.insert.subiekt-gt.yml
```diff
@@ -0,0 +1,43 @@
...
+  date:
+    parser: regex
+    regex:
+      - Data wystawienia:\n.*(\d{2}\.\d{2}\.\d{4})
+      - Data wystawienia:\n.*(\d{4}-\d{2}-\d{2})
+    type: date
+  invoice_number:
+    parser: regex
+    regex: Faktura VAT\s+(.*?)\s+oryginał
+  amount:
```
bosd, line 30: The template is caputing the amount. Is it also possible to capture:
amount_untaxed (The total amount of the invoice (excluding taxes))
amount_tax (The sum of the tax amount of the invoice)

src/invoice2data/extract/templates/pl/pl.insert.subiekt-gt.yml
```diff
@@ -0,0 +1,43 @@
...
+    parser: regex
+    regex: Razem do zapłaty:\s+([\d\s]+,[\d][\d])
+    type: float
+  nrb:
```
bosd, line 34: Would it be possible to also capture the `iban` from these invoices?

bosd (changes requested):
I'm about to give this one a :+1: as tests are all green.
Please attend to comments if possible.

src/invoice2data/extract/templates/pl/pl.insert.subiekt-gt.yml
bosd: We can do this later on in an update of this template or code. Not worthwhile blocking this PR for that.

src/invoice2data/extract/templates/pl/pl.insert.subiekt-gt.yml
bosd: We can do this later on in an update of this template. Not worthwhile blocking this PR for that.

src/invoice2data/extract/templates/pl/pl.insert.subiekt-gt.yml
bosd: Personally don't have the knowledge if this can be replaced by vat.
Is it a personal or business vat number being parsed??

Currently not worthwhile blocking the pr for this.
When disered updates on the template can be made later on.

bosd (approved):
"Fasttracking" this as getting the functionality is more of an priority then debating my detailed comments.

bosd:
Made the same performance update as on the 'old code' in #470
Tests will pass after the flipkart template has been updated from that pr.

bosd:
btw, I run this trough the profiler. No noticable difference on the performance.

bosd:
Just reverted the performance changes. As we are discussing them in #470.
This one is now independent from that one.
When tests are green, we can merge this one.

PR #417, opened 2022-10-01
Lines: support multiple RegEx-es in lines entries

rmilecki:
```
So far "first_line", "line" and "last_line" could contain a single
RegEx only. Some invoices have lines that use more than one format. To
simplify parsin them allow all 3 entries to contain list of RegEx-es.

Example:
fields:
  lines:
    parser: lines
    start: Item\s+Discount\s+Price$
    end: \s+Total
    line:
      - Items group:\s+(?P<group>.+)
      - (?P<description>.+)\s+(?P<discount>\d+.\d+)\s+(?P<price>\d+\d+)
```

PR #422, opened 2022-10-08
templates: de: improve parsing QualityHosting lines

rmilecki:
```
A single QualityHosting invoice position spans across multiple lines.
For that reason it uses a very generic RegEx for middle lines:
line: '^\s+(?P<desc>.+)$'

That doesn't work well with multi-page invoices. It's because above
RegEx matches page footer lines. That results in footer content getting
extracted as invoice line "desc".

Improve that situation by adding "last_line" RegEx matching position
last line. That prevents parsing lines between last and first lines
(e.g. footer content).
```

### DIFF
--- a/src/invoice2data/extract/templates/de/de.qualityhosting.yml
+++ b/src/invoice2data/extract/templates/de/de.qualityhosting.yml
@@ -12,6 +12,7 @@ lines:
   end: 'Total EUR'
   first_line: '\s+(?P<pos>\d+)\s+(?P<qty>\d+)\s+(?P<desc>.{,70})\s+(?P<price>\d+,\d+)'
   line: '^\s+(?P<desc>.+)$'
+  last_line: '^\s+(?P<desc>\d\d\.\d\d\.\d\d-\d\d\.\d\d\.\d\d)$'
   types:
       qty: float
       price: float
--- a/tests/compare/QualityHosting.json
+++ b/tests/compare/QualityHosting.json
@@ -41,13 +41,13 @@
             {
                 "pos": "6",
                 "qty": 1.0,
-                "desc": "Small Business QualityExchange 2010\nGrundgebühr pro Einheit\nDienst: OUDJQ_jauernik\n01.05.14-31.05.14\nQualityHosting AG - Uferweg 40-42 - D-63571 Gelnhausen\niViveLabs Ltd.\n93B Sai Yu Chung\nYuen Long, N.T.\nHong Kong\nPos.            Menge      Beschreibung                                                            Rabatt %     VK-Preis     Zeilenbetrag\nOhne      Ohne MwSt.\nMwSt.",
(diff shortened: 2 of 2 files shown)


rmilecki:
So what do we do about this pull request?

It received some minor cross-pull-request comment in the #417:

> Maybe better to leave suboptimal tests and examples in this library. Just as an showcase. (Same goes for the OCR examples in this repo). It is definityle helping us to find these corner cases.

As `de.qualityhosting.yml` is a template for actual (real-life) invoices I really think we should fix it. By that I mean accepting this pull request.

If we need to test some corner cases - that can be done using custom templates & tests. I added support for such in the just-merged #414.

bosd:
> So what do we do about this pull request?

I'd suggest we let this one sit here for a moment.
At least until we've sorted out the multiline parsing.

Wen't trought the re lib docs. need some time to test and verify things. Will report back in #417

bosd:
As in 417 mentioned.
I agree to update this template.
Possibly to include only lines with more then 30 spaces in front
`^\s{30,}`

For this particular invoice we could make the lastline specific, (or either leave it very generic, like before this pr)
Just the question wat kind of example we would like to include.

As per your previous suggestion:
`'^\s+(?P<desc>\d\d\.\d\d\.\d\d-\d\d\.\d\d\.\d\d)$'`
I would like to propose to change it to
`'^\s{30,}(?P<desc>\d{2}[.]\d{2}[.]\d{2}[.]-\d{2}[.]\d{2}[.]\d{2})$'`
rewrote the `.` to `[.]` as it is best practice to prevent the use of a plain `.` preformance wise.

(I quikly drafted this, it needs testing..)

bosd:
Used this version of thetemplate to check the code of #417.
Intentional it is without `last_line`. Because the purpose of the check was to see if it is adding a matched line to the output without the lastline key.

```
issuer: QualityHosting AG
fields:
  amount: Total EUR\s+(\d+,\d+)
  amount_untaxed: Total EUR\s+(\d+,\d+)
  date:
    - \s{2,}(\d+\. .+ \d{4})\s{2,}
    - Rechnungsdatum\s+(\w+ \d+, \d{4})
  invoice_number: Rechnungsnr\.\s+(\d{8})
  vat: DE 232 446 240
lines:
  start: 'Contract No. \w+'
  end: 'Total EUR'
  first_line: '\s+(?P<pos>\d+)\s+(?P<qty>\d+)\s+(?P<desc>.{,70})\s+(?P<price>\d+,\d+)'
  line: '^\s{30}(?P<desc>.{5,30})$'
  types:
      qty: float
      price: float
keywords:
- QualityHosting
options:
  currency: EUR
  decimal_separator: ","

```  

@rmilecki  Is it ok for you to use this version?

bosd:
Rebased onto current master (the comparison fixture had since been reformatted) and applied as `9a5c31e`; the QualityHosting expected output was re-derived from the actual parser result and verified by the test suite. Shipped in **v0.4.6** (https://pypi.org/project/invoice2data/0.4.6/). Thanks @rmilecki! Closing as released.

PR #424, opened 2022-10-21
Added com.newrelic.yml template

esmeraldas63:


bosd:
Thanks for this contribution. 

A bit of nitpikking, but would suggest to wrap the `.` as the decimal/date separator in brackets like `[.]` as it is an exact match.
(It's better performance wise)

I don't have an invoice to test it against. But I would assume you already did that.
I'm not from the usa, so don't know the situation.
but this invoice has an tax id, but no tax amount is caputured is that correct?

rmilecki:
> A bit of nitpikking, but would suggest to wrap the `.` as the decimal/date separator in brackets like `[.]` as it is an exact match.
> (It's better performance wise)

I agree, though my personal suggestion & preference would be to use `\.`. You're matching a dot so I think it makes sense to use a dot (`\.`) regex.

esmeraldas63:
Good catch, thanks, actually date separator is "/". Updated regexp to match exactly that.

paveikslas

esmeraldas63:
> I don't have an invoice to test it against. But I would assume you already did that. I'm not from the usa, so don't know the situation. but this invoice has an tax id, but no tax amount is caputured is that correct?

Added regexp to capture tax_amount :)

bosd (approved):

PR #429, opened 2022-10-24
Update doc, start used by list

bosd:
Start a list of projects who use this library.
Usefull for the end user as an overview which gui's use this lib.
Usefull for developers to check, to avoid breaking changes.


rmilecki:
I'm neutral on this.

PR #438, opened 2022-11-28
Added area support for fields (with unit test)

alexis-via:


alexis-via:
supersedes PR https://github.com/invoice-x/invoice2data/pull/333

src/invoice2data/extract/invoice_template.py
```diff
@@ -151,12 +154,23 @@ def coerce_type(self, value, target_type):
             return self.parse_date(value)
         assert False, "Unknown type"
 
-    def extract(self, optimized_str):
+    def extract(self, optimized_str: str, invoice_file: str, input_module: str) -> Optional[dict]:
         """
         Given a template file and a string, extract matching data fields.
-        """
 
-        logger.debug("START optimized_str ========================\n" + optimized_str)
```
bosd, line 159: Could you leave the debugger output i this form? It prevents the first line from being misaligned.

bosd (changes requested):
Could you leave the debugger output i this form? It prevents the first line from being misaligned.

src/invoice2data/extract/invoice_template.py
```diff
@@ -174,11 +188,33 @@ def extract(self, optimized_str):
         output["issuer"] = self["issuer"]
 
         for k, v in self["fields"].items():
+            # k is the key of the field
+            # v is the value
             if isinstance(v, dict):
+                # Options were supplied to this field
+                if "area" in v and input_module == pdftotext:
```
bosd, line 195: The Tesseract input module is using pdftotext now too. To format the output with the position of the original text.
https://github.com/invoice-x/invoice2data/blob/4d517d131fdaa568b4282f2f0a12abef98751075/src/invoice2data/input/tesseract.py#L100-L118

Does it make sense to add tesseract as an option, to also use the area support?

bosd (approved):
LGTM :+1: 
The other comment is an optional enhancement. Not blocking for this commit.

bosd:
@alexis-via I see you prepared for the tesseract module as well.

For the tesseract module to support area field. The changes made to the pdftotext.py needs also be applied to the tesseract.py file.
Is that something you would like to include in this PR?

Otherwise we can merge this as is.
(pr's for this functionality were already open for a long time)

Issue #109, opened 2018-04-18
Tesseract: AttributeError: 'str' object has no attribute 'decode'


RobertLemmens:
Ive been trying out the tesseract ability but i keep getting this particular error. I see it go through page 1 and then page 2 and then it throws the following:

Traceback (most recent call last):
  File "/usr/bin/invoice2data", line 11, in <module>
    load_entry_point('invoice2data==0.2.81', 'console_scripts', 'invoice2data')()
  File "/usr/lib/python3.6/site-packages/invoice2data/main.py", line 117, in main
    res = extract_data(f.name, templates=templates, input_module=input_module)
  File "/usr/lib/python3.6/site-packages/invoice2data/main.py", line 43, in extract_data
    extracted_str = input_module.to_text(invoicefile).decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'

Im not quite sure what the error means in this context. I tried converting the pdf into a tiff myself as defined in the tesseract.py file, and then running tesseract standalone. tesseract creates output as expected, so i believe its something in the application and not tesseract.

RobertLemmens:
Quick update on this; i downloaded the source and removed .decode('utf-8') on line 13 in tesseract.py because it seems that main.py already does this and now it compiles fine. I dont usually work with python but ill put in a PR so you can look at what i did. Im still not sure if its something on my end or with the framework

blade3609:
I had to do the exact same thing, now it works in python 2.7 and Python 3.5

m3nu:
Good find @RobertLemmens . I believe someone only changed this recently. I'll check why this wasn't covered by tests.

m3nu:
Merged your PR. The Tesseract module is not very frequently used and we don't test it yet. Hoping to see some improvements here over the summer.

duskybomb:
`.decode('utf-8')` was added because to fix the problems which occur in Windows.
See #32 and #99

duskybomb:
Tesseract module is not working well. There were simple issues such as 
`$101.00` was being read as `$101. 00` and hence says some fields are not matching, therefore not much point in making tests at this point (almost zero usability). I was reading more about tesseract module and they have released v4.0(beta) on 28 March 2018, I hope it works better.

bosd:
Since the time this issue has been opened there has been a lot of work done on both tesseract and invoice2data.
Right now it is possible to get some great usable results.

However the results are not 100% consistent.
[...]

PR #439, opened 2022-11-30
Tmpl nl mobility

bosd:
With these templates you will be able to process 95% of dutch fuel station receipts.
Additionnaly:
- ANWB Road assistance
- ipparking
- ns rail invoices
- parkmobile parking app

### DIFF
--- a/src/invoice2data/extract/templates/be/be.eg_retail.yml
+++ b/src/invoice2data/extract/templates/be/be.eg_retail.yml
@@ -1,0 +1,87 @@
+# SPDX-License-Identifier: MIT
+issuer: EG Retail (Belgium) BVBA
+fields:
+  amount:
+    parser: regex
+    regex:
+      - (?i)TOTAAL\s+€?.?\s+(\d+.\d{2})
+      - (?i)Amex\s+€.?\s+(\d{1,3}.\d{2})
+    type: float
+  amount_untaxed:
+    parser: regex
+    regex:
+      - \d{2}.\d{2}\s+. (\d+.\d{2})\s+. \d+.\d{2}
+    type: float
+    group: sum
+  date:
+    parser: regex
+    regex: (\d{1,2}-\d{2}-\d{4}\s+\d{1,2}.\d{2}.\d{2})
+    type: date
+  invoice_number:
+    parser: regex
+    regex: (\d{1,2}-\d{2}-\d{4})\s+\d{1,2}.\d{2}.\d{2}
+  static_vat: BE04060843239
+  partner_name:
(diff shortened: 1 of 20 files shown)


rmilecki:
What's the point of that commented out code repeated in so many templates?
```
# below code is wrong pump name and with product field
#    line: .(?P<product>\w*).?\s+.\s.\s+\d+.\d{0,2}\n.?\s+.(?P<name>POMP.*) (?P<qty>\d+.\d{2})\s. .\s+€ (?P<price_unit>\d.\d{3})
# Above one modified for file without line breaks
#    line: .(?P<product>\w*).?\s+.\s.\s+\d+.\d{0,2}.*.?\s+.(?P<name>POMP.*) (?P<qty>\d+.\d{2})\s. .\s+€ (?P<price_unit>\d.\d{3})
# same as above optimized for (non optimized string)
#    line: .(?P<product>(\w+(?:\S|[.]\w\w+|\n)*)).?\s+.\s.\s+\d+.\d{0,2}.*.?\s+.(?P<name>POMP.*) (?P<qty>\d+.\d{2})\s. .\s+€ (?P<price_unit>\d.\d{3})
```

bosd:
> What's the point of that commented out code repeated in so many templates?

I'll have a look at it again.
I've started making these templates a while ago. Back then I assumed that 378 got merged.
But that PR did not get merged. There have been some alternative pr's since that one.

Need to look at it again, if the alternative pr's are now providing the same functionality.

bosd:
Interesting to add..
Some of these templates are for recognizing tickets from POS systems.
That is a particular challenging task. Because the receipts need to be scanned and processed by tesseract ocr.
I am able to achieve some impressive results. Really usable. So I am willing to provide these as an example here.
However, the results are still inconsistent.

By that I mean, with the same input file, same processing command, the result can be different.
Therefore not suitable for using them as an exact unit test match.


Currently tesseract is untested see #109 .
Since then tesseract has improved a lot. It is usable, we can try to make a unittest for it.
Maybe under ideal conditions we could get 100% repeatative results..
Alhough I am not sure.
Is there a way to add a non blocking test.?
So when it fails, it would not block the whole github actions cycle?

bosd:
I appoligize for all the activity in this PR.
I tried to fix the future warning. But it was not introduced by these commits.
Eventually found it, and fix is in #516 

This PR has been sitting here for quite some while.
It is one of the items I want to get off my task list.
So, Will make this ready to be merged.

If one of these templates needs update. I will do that in micro commits / PR's.
Going to sqaush all the commits here. And merge this one.

PR #440, opened 2022-12-08
Area support for tesseract

bosd:
This allows one to specify the page(s) and coordinates of the area to be extracted.


bosd:
Found a little mistake, ready for review now

rmilecki (approved):

PR #446, opened 2023-01-10
Add template encoding option

### LINKED ISSUE #445: Automatic encoding detection for template files fails for single german umlaut
In some cases `chardet.detect` fails to detect the correct template encoding. This leads to wrong `InvoiceTemplate` objects and wrong parsing results in the end.

Example template:
```yaml
issuer: Basic Test
keywords:
  - Basic Test
fields:
  date:
    parser: regex
    regex: Issue date:\s*(\d{4}-\d{2}-\d{2})
    type: date
  invoice_number:
    parser: regex
    regex: Invoice number:\s*([\d/]+)
  amount:I have implemented a new `option` to specify the actual template encoding and extended the loader mechanism
    parser: regex
    regex: Total:\s*(\d+\.\d\d)
    type: float
  single_specialchar:
    parser: static
    value: ä
```

If I add another umlaut the detection works and identifies as `utf-8`. As a quick workaround one could add another umlaut to an unused static value.

timorieber:
Fixes #445.

I have implemented a new `option` to specify the actual template encoding and extended the loader mechanism. Also did some refactoring in the `read_templates` function for better readability and testability.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -58,11 +58,11 @@ All the regex `fields` you need extracted. Required fields are `amount`,
 `date`, `invoice_number`. It's up to you, if you need more fields
 extracted. Each field can be defined as:
 
-- an **associative array** with 
-`parser` (required) specifying parsing method and 
-`area` (optional) specifying the region of the pdf to search. 
-This takes the following arguments: `f` (first page), `l` (last page), `x` (top-left x-coord), `y` (top-left y-coord), 
-`r` (resolution), `W` (width in pixels) and `H` (height in pixels). When setting your region, ensure the resolution in your 
+- an **associative array** with
+`parser` (required) specifying parsing method and
+`area` (optional) specifying the region of the pdf to search.
+This takes the following arguments: `f` (first page), `l` (last page), `x` (top-left x-coord), `y` (top-left y-coord),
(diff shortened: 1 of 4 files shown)


rmilecki:
I'm not sure if we need all that encoding at all. We should also clarify that unclear
```
# -*- coding: utf-8 -*-
```
like syntax.

As for you actual changes here, your commit "Implement yaml option for template encoding" does a lot of refactoring and rewording that isn't related to actual issue. In my opinion that should go to another pull request so we can focus on one thing at a time.

timorieber:
This PR is meant as a short-term backwards-compatible approach to this template encoding issue. For the longer term I support your plan of removing this "magic" and assume `utf-8` template encoding.

I had a harder time than expected understanding what's going on in this code. As I'm a big fan of Robert C. Martin's "boyscout rule" to "leave the code better than you found it" and Martin Fowler's "opportunistic refactoring" I first wrote some tests (there were none) and afterwards split things up a little bit. I'm pretty confident that it works just like expected.

I'm also fine when this PR gets closed, but I think it's the right direction.

rmilecki:
Obsoleted by https://github.com/invoice-x/invoice2data/pull/449

Cleanup of existing templates is pending: https://github.com/invoice-x/invoice2data/pull/452

Issue #442, opened 2022-12-19
Failed to load nl.begra.yml template


MKatereggaExigent:
'''
Failed to load nl.begra.yml template:
while parsing a block mapping
  in "<unicode string>", line 5, column 3:
      amount: Totaal EUR incl[.] btw\s ...
      ^
expected <block end>, but found '{'
  in "<unicode string>", line 12, column 14:
      iban: [A-Z]{2}\d{2}?\w{4}?\d{4}?\d{4}?\d{0,2}
                 ^
Loaded 164 templates from /Users/<USER>/opt/anaconda3/lib/python3.9/site-packages/invoice2data/extract/templates
No template for data/my_Invoice.pdf
'''

rmilecki:
Fixed by 43778de24e7a ("Fix nl.begra.yml formatting")

PR #447, opened 2023-01-21
Fix nl.begra.yml formatting

rmilecki:
```
This fixes:

Failed to load nl.begra.yml template:
while parsing a block mapping
  in "<unicode string>", line 5, column 3:
      amount: Totaal EUR incl[.] btw\s ...
      ^
expected <block end>, but found '{'
  in "<unicode string>", line 12, column 14:
      iban: [A-Z]{2}\d{2}?\w{4}?\d{4}?\d{4}?\d{0,2}
                 ^
```


bosd (approved):
Code review, LGTM

bosd:
This should fix #442

Issue #448, opened 2023-01-21
UTF-8 template file with 1 Polish character parsed using ISO-8859-1


rmilecki:
I'm creating this issue to provide another example of unreliable `chardet` behaviour.

A following simple template:
```
# -*- coding: utf-8 -*-
# SPDX-License-Identifier: MIT
issuer: PayPro
keywords:
  - 'NIP: PL779-236-98-87'
fields:
  vatin:
    parser: static
    value: PL7792369887
  amount: Do zap.aty:\s+(\d[\d\s]*\.\d{2})
  date: Data wystawienia:\s+(\d{4}-\d{2}-\d{2})
  sale_date: Data sprzedaży:\s+(\d{4}-\d{2}-\d{2})
  invoice_number: NR:\s+([\dA-Z/]+)
options:
  currency: PLN
  date_formats:
    - '%Y-%m-%d'
  decimal_separator: '.'
```
gets encoding incorrectly detected as ISO-8859-1 (instead of utf-8) and its `sale_date` RegEx doesn't get matched against invoice content.

rmilecki:
Solved by 6ab31047386a ("Treat all YAML templates files as UTF-8 encoded")

PR #449, opened 2023-01-21
Treat all YAML templates files as UTF-8 encoded

rmilecki:
```
Treat all YAML templates files as UTF-8 encoded

Encoding detection based on chardet is unreliable. For UTF-8 encoded
files with few non-ASCII chars it often fails to detect UTF-8. That
results in invoices not being parsed correctly.

Read every YAML file as UTF-8 encoded. It is used by all project
internal templates. It should be used by all external ones too.

FWIW YAML documentation also says that every *processor* must support
the UTF-8, UTF-16 and UTF-32:
https://yaml.org/spec/1.2.2/#52-character-encodings
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,6 @@ setup_requires =
   pytest-runner
   setuptools_git
 install_requires =
-  chardet
   pillow
   pyyaml
   dateparser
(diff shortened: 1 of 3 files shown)


rmilecki:
This fixes:
* https://github.com/invoice-x/invoice2data/issues/368
* https://github.com/invoice-x/invoice2data/issues/445
* https://github.com/invoice-x/invoice2data/issues/448

It's an alternative solution to the https://github.com/invoice-x/invoice2data/pull/446 . I'm not sure we should / have to support any other encoding.

bosd (approved):
Code review, LGTM

Issue #441, opened 2022-12-12
Error "AttributeError: 'list' object has no attribute 'strftime'" when using "_date" in field name


iedev1:
When defining fields for extraction if the field name is either prefixed with `date_` or suffixed with `_date`, invoice2data errors with an `AttributeError`.

This happens regardless of setting the `type` property to date/string.

If I simply rename the field, the extraction/capturing works as expected.

Here is the relevant line from the config:

```
fields:
    # [...]
    deposit_date:
        parser: regex
        regex: 
          - Deposit Date: (\d{2}\/\d{2}\/\d{4})
        type: date
```

Here is the error:

```
Traceback (most recent call last):
  File "/opt/homebrew/bin/invoice2data", line 8, in <module>
    sys.exit(main())
  File "/opt/homebrew/lib/python3.10/site-packages/invoice2data/main.py", line 231, in main
    output_module.write_to_file(output, args.output_name, args.output_date_format)
  File "/opt/homebrew/lib/python3.10/site-packages/invoice2data/output/to_json.py", line 46, in write_to_file
    line[k] = v.strftime(date_format)
AttributeError: 'list' object has no attribute 'strftime'
```

bosd:
A couple of minutes after this ticket was openend an new version has been published on pypi.
Which version are you using??
Could you update to the latest version and check if it is resolved?

rmilecki:
I tried to reproduce this but I actually got yet another problem:
```
Traceback (most recent call last):
  File "/home/foo/.local/bin/invoice2data", line 11, in <module>
    load_entry_point('invoice2data', 'console_scripts', 'invoice2data')()
  File "/home/foo/invoice2data/src/invoice2data/main.py", line 224, in main
    res = extract_data(f.name, templates=templates, input_module=input_module)
  File "/home/foo/invoice2data/src/invoice2data/main.py", line 108, in extract_data
    return t.extract(optimized_str, invoicefile, input_module)
  File "/home/foo/invoice2data/src/invoice2data/extract/invoice_template.py", line 216, in extract
    value = parser.parse(self, k, v, optimized_str_for_parser)
  File "/home/foo/invoice2data/src/invoice2data/extract/parsers/regex.py", line 33, in parse
    matches = re.findall(regex, content)
  File "/usr/lib64/python3.6/re.py", line 222, in findall
    return _compile(pattern, flags).findall(string)
  File "/usr/lib64/python3.6/re.py", line 289, in _compile
[...]

PR #450, opened 2023-01-22
parsers: regex: check if regex from template is a string

rmilecki:
```
Incorrectly formatted templates may result in passing OrderedDict as a
regex. That makes re.findall fail like:

  File "/home/foo/invoice2data/src/invoice2data/extract/parsers/regex.py", line 33, in parse
    matches = re.findall(regex, content)
  File "/usr/lib64/python3.6/re.py", line 222, in findall
    return _compile(pattern, flags).findall(string)
  File "/usr/lib64/python3.6/re.py", line 289, in _compile
    p, loc = _cache[type(pattern), pattern, flags]
TypeError: unhashable type: 'collections.OrderedDict'

Handle such case by displaying a meaningful warning.
```

### DIFF
--- a/src/invoice2data/extract/parsers/regex.py
+++ b/src/invoice2data/extract/parsers/regex.py
@@ -30,6 +30,9 @@ def parse(template, field, settings, content, legacy=False):
 
     result = []
     for regex in regexes:
+        if not isinstance(regex, str):
+            logger.warning("Field \"%s\" regex is not a string (%s)", field, str(regex))
+            continue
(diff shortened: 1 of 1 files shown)


rmilecki:
I experienced this while trying to debug https://github.com/invoice-x/invoice2data/issues/441

bosd (approved):
Lgtm

rmilecki:
This sould also help with https://github.com/invoice-x/invoice2data/issues/339

PR #451, opened 2023-01-22
output: to_json: improve formatting dict-ionaries

rmilecki:
```
Previous code was:
1. Expecting every field with "date" in its name to be datetime.date
2. Ignoring datetime.date in other fields

That didn't match possible parsing results after all recent changes:
1. Any field can contain a date
2. Fields with "date" in their names can be lists

This change rewrites to_json to allow dealing with datetime.date in any
field at any depth level.

This also fixes possible issue like:

    File "/opt/homebrew/bin/invoice2data", line 8, in <module>
    sys.exit(main())
    File "/opt/homebrew/lib/python3.10/site-packages/invoice2data/main.py", line 231, in main
    output_module.write_to_file(output, args.output_name, args.output_date_format)
    File "/opt/homebrew/lib/python3.10/site-packages/invoice2data/output/to_json.py", line 46, in write_to_file
    line[k] = v.strftime(date_format)
AttributeError: 'list' object has no attribute 'strftime'
```

### DIFF
--- a/src/invoice2data/output/to_json.py
+++ b/src/invoice2data/output/to_json.py
@@ -3,10 +3,18 @@ import datetime
 import codecs
 
 
-def myconverter(o):
-    """function to serialise datetime"""
-    if isinstance(o, datetime.datetime):
-        return o.__str__()
+def format_item(item, date_format):
+    if isinstance(item, datetime.date):
+        return item.strftime(date_format)
+    elif isinstance(item, datetime.date):
+        return item.__str__()
+    elif isinstance(item, dict):
+        for k, v in item.items():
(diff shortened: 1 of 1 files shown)


rmilecki:
This fixes https://github.com/invoice-x/invoice2data/issues/441

bosd (approved):
LGTM

bosd:
I'm using this as a lib. Also to digitize receipts.
There are cases where I have multiple receipts from the same issuer on a day.
With the regex I am capturing the date including time. The return is a `datetime` instance. 

Is it an idea to make date time instance the default? Or how to differentiate between `datetime` and `date`

PR #452, opened 2023-01-22
templates: drop unused coding info from all templates

rmilecki:
```
For some unkown reason early templates contained line like:
    # -*- coding: utf-8 -*-

It seems to be unused. It doesn't belong to YAML specification. It seems
like a copy & paste mistake from Python source file.

Remove all those unused lines. All YAML templates as expected to be
UTF-8 encoded anyway.

Those changes were automatically generated using:
sed -i '/coding:/d' src/invoice2data/extract/templates/*/*.yml
find ./tests/custom/ -name "*.yml" -exec sed -i '/coding:/d' {} \;
```

### DIFF
--- a/src/invoice2data/extract/templates/ch/ch.pcengines.yml
+++ b/src/invoice2data/extract/templates/ch/ch.pcengines.yml
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 issuer: PC Engines GmbH
 fields:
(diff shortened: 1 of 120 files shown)


rmilecki:
Discovered in the https://github.com/invoice-x/invoice2data/issues/445#issuecomment-1399293692

bosd (approved):

Issue #297, opened 2020-08-07
UnboundLocalError: local variable 'first_row' referenced before assignment


ShaileshSarda:
I am getting this: UnboundLocalError: local variable 'first_row' referenced before assignment
error to extract data in csv format.

Any help?
Thanks!

m3nu:
Maybe you didn't get any results? Else it's hard to tell without full stack trace.

PR #453, opened 2023-01-22
output: to_csv: print header for every set of parsed fields

rmilecki:
```
When parsing multiple invoices it's *incorrect* to:
1. Display header with fields names of the *first* parsed invoice
2. Display content of *all* parsed invoices

It's because parsing dfferent invoices may result in different fields.
In such case header won't make sense for all rows.

To improve handling such situations display new header for every set of
fields that differs from the last one.
```

### DIFF
--- a/src/invoice2data/output/to_csv.py
+++ b/src/invoice2data/output/to_csv.py
@@ -41,13 +41,16 @@ def write_to_file(data: list, path: str, date_format="%Y-%m-%d") -> None:
     with openfile as csv_file:
         writer = csv.writer(csv_file, delimiter=",")
 
+        last_header = None
         for line in data:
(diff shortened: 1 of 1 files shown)


rmilecki:
This also fixes https://github.com/invoice-x/invoice2data/issues/297

rmilecki:
Ping @bosd :)

bosd (approved):
I'm neutral on this. (Don't use csv output) Code LGTM

rmilecki:
Me neither, I don't think CSV is a good fit for tree structures anyway. I just wanted to fix that issue and noticed an extra problem with doing so.

Thank you for looking at this pull request!

PR #456, opened 2023-01-28
Don't set empty fields in output

rmilecki:
```
Before introducing parsers concept extract() did not set empty fields in
output. So in case of re.findall() returning an empty list there was a
warning like:
WARNING:invoice2data.extract.invoice_template:regexp for field <foo> didn't match
and on <foo> in output dictionary / file.

That has unintentionally changed after commit 670f8729b7b6 ("Extend
fields syntax to support associative arrays & add 2 initial parsers
(#308)"). Output can now look like:
[
    {
        "issuer": "Company",
        "amount": 12.34,
        "date": "2023-01-01",
        "invoice_number": "01/2023",
        "test_field": [],
        "currency": "EUR",
        "desc": "Invoice from Company"
    }
]

Bring back old behaviour and don't set empty fields:
[
    {
        "issuer": "Company",
        "amount": 12.34,
        "date": "2023-01-01",
        "invoice_number": "01/2023",
        "currency": "EUR",
        "desc": "Invoice from Company"
    }
]
```

### DIFF
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -214,7 +214,7 @@ class InvoiceTemplate(OrderedDict):
                     if v["parser"] in PARSERS_MAPPING:
                         parser = PARSERS_MAPPING[v["parser"]]
                         value = parser.parse(self, k, v, optimized_str_for_parser)
-                        if value is not None:
+                        if value:
                             output[k] = value
                         else:
                             logger.error("Failed to parse field %s with parser %s", k, v["parser"])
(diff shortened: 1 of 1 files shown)


rmilecki:
I've just noticed that my old https://github.com/invoice-x/invoice2data/pull/308 changed parsing behaviour.

I'm wondering if the old behaviour made more sense and should be brought back. Any opinion on this?

bosd:
I am kinda neutral on this. Both methods has his up and downsides.
But I tend to lean towards returning to the old behavior.

bosd (approved):
Let's restore the old behavior. :+1

PR #457, opened 2023-01-29
Refactor and test template loader mechanism

timorieber:
Thanks for your work. It was a good decision to drop encoding detection and simply treat all templatefiles as `utf-8`. This made my previous PR #446 also obsolete.

As I put some effort in improving the loader mechanism and make it easier to read and test, I propose these changes again without encoding detection.

I would have enhanced it further (e.g. using `pathlib.Path`), but the Python 2.7 compatibility stopped me. Are there plans to only support actively maintained Python versions in upcoming releases? I could help at that, too.

### DIFF
--- a/src/invoice2data/extract/loader.py
+++ b/src/invoice2data/extract/loader.py
@@ -1,7 +1,7 @@
 import codecs
 import logging
-import os
 from collections import OrderedDict
+from pathlib import Path
 
 import pkg_resources
 import yaml
@@ -27,7 +27,7 @@ def ordered_load(stream):
     return yaml.load(stream, OrderedLoader)
 
 
-def read_templates(folder=None):
+def read_templates(folder: str = None):
     """
     Load yaml templates from template folder. Use built-in templates if no folder is set.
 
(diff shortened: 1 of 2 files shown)


bosd:
Thanks for this contribution. Looking at the template loader was on my todo.
I've got the impression that there are some possibilities to improve the performance of the loading of templates.
Does this PR improve (or impact) performance??

Before changing the template loader system. I would prefer to add another tests. This library is used in cases where the templates are loaded from another db and provided as a stream. (Is that the correct description?) 
I would like to include this as a test, just to make sure, that implementation won't break.

Will dig later into these changes. For now, I will activate the github workflow tests.

rmilecki:
I took some early attempt on reviewing this but it was trivial, there are a lot of changes. Also some unrelated changes (changing code order) make it more difficult than needed. I'll have to give it another try.

timorieber:
> Does this PR improve (or impact) performance??

No. It uses all the same routines as before in the same order, only split up in more separated functions with less responsibilities.

> Also some unrelated changes (changing code order) make it more difficult than needed.

I've simplified the implementation around the `OrderedDict` and put it within the only function relying on it, to make it more clear that it is only an internal helper mechanism that could be replaced easily in future work.

Also I use some basic auto-formatting in my IDEs,  because in my projects there is much more code standardisation applied in later stages, e.g. local tox runs or CI builds. Here this automatism affects the sorted imports.

Sorry, that this PR is more complicated than it perhaps needed to be. Some more commits in between could have improved the review and would explicitly show the steps I have taken.

bosd:
> I would have enhanced it further (e.g. using `pathlib.Path`), but the Python 2.7 compatibility stopped me. Are there plans to only support actively maintained Python versions in upcoming releases? I could help at that, too.

Python 2.7 seems to be unsupported for 3 years now. We are not actively testing it in the workflow.
I think it is time to remove the python 2.7 label from the documentation. And let developers won't hold back their commits due to compatibility.

bosd (changes requested):
Thanks again for this pr.
Did some extensive testing!

This weekend I was playing with a profiler, to find out why this lib is so slow.
More on that in different PR's.

I've tested this code.. 
The readability is a question of personal preference. (I like a bit of splitting in several functions :+1: )
The negative thing of the splitting is performance. :sob: 
The number of function calls is exploding with this code.
(as each loaded template will have multiple calls)
And function calls are quite expensive in python performance wise.
Resulting in the code being 1,5x times as slow in the profiler.

Testing on a single invoice.
old code:  **2579** function calls
this pr: **293413** function calls

So my vote is to not merge this pr in the current state.
Rather use it as an inspiration.
- Update variable names
- format logger strings
- tests for the loader.

FYI:
I've also made an intermediate version, of this code with the old code.
(Less splitting of functions) But still I came to the conclusion that performance is greatly impacted by this. So better not to split the loader functions.

timorieber:
Hi @bosd,

based on your feedback I gave it another try. All changes are encapsulated in small and reviewable commits to easily follow my iterative thaughts for optimisation.

The loader mechanism is now kept in a single function and thanks to dropped support for Python 2.7 I was able to use simpler approaches, e.g. `pathlib.Path`, which made things a lot easier to implement. I didn't measure, but because of this fact, I'm pretty confident the performance has improved at least a little bit, too.

The function is now much cleaner and easy to understand, also the key assumptions are properly tested.

bosd:
Hi @timorieber 
I forgot to mention I was working on a competing pr. It is in #471 
In that one I took into account some performance measures:

- Usage of f strings in the logger are 8% slower then %s format. So better to stick to the (old) %s format for now.
- Pathlib is newer, yet slower then the old method. So did not change that.
- Ordereddict seems to be a leftover from python 2 days. It seems it is not really required anymore. 
OrderedDict is over 80% slower than the standard Python dictionary.
source: https://python.plainenglish.io/should-we-still-use-ordereddict-in-python-f223c85a01d5

There still are some very usefull improvements here.
Will have to look how to continue from here. As the 471 pr is already a large diff.

timorieber:
Thanks for the feedback, @bosd. I'll close this for now.

Issue #155, opened 2018-08-04
No warning with lines


duskybomb:
If regex defined for a line is not matched, no warning is shown unlike with other regular fields.

rmilecki:
Fixed by #460

PR #460, opened 2023-02-03
parsers: lines: display "ignoring (...)" for all unrecognized lines

rmilecki:
```
This is very helpful for debugging problems with parsing lines.
```

### DIFF
--- a/src/invoice2data/extract/parsers/lines.py
+++ b/src/invoice2data/extract/parsers/lines.py
@@ -63,45 +63,43 @@ def parse_block(template, field, settings, content):
                 # This will allow last_line and line to be matched on below
                 first_line_found = True
                 continue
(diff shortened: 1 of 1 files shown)


rmilecki:
This addresses:

- https://github.com/invoice-x/invoice2data/issues/155
- https://github.com/invoice-x/invoice2data/issues/339

This diff looks quite big, but the real change is `if first_line_found is False:` → `if first_line_found is True:` and comment update. The rest of diff lines are whitespace changes (indention level).

To review this pull request easily please use `git`'s option `-w` or GitHub's setting of "Diff view" called "Hide whitespace"

rmilecki:
See actual changes in this pull request:
image

bosd (approved):
Was afraid this would affect the results.
Tests are still passing, so LGTM

PR #428, opened 2022-10-23
Example coolblue

bosd:
These invoices have 2 different types of lines.

The problem is in the line
`Incl. Thuiskopieheffing: Thuiskopie €3.50 1 21% € 4,24`
where the content of the column `prijs per stuk` has moved to the front.

Made a template in the past, which could parse both the regular lines and the custom line as mentioned above.
```
lines:
 - start: 'Artikel'
   end: 'Exclusief BTW'
   first_line: ^(?P<product>(\w+(?:\s\S+)*))\s+(?P<qty>[-]?\d+)\s+[€]?\s+(?P<price_unit>(\d{,3}[.]?\d{,3}[,]\d{2}))\s+(?P<tax_percent>\d+)[%]?\s+[€]?\s(?P<price_subtotal>([-]?\d{,3}[.]?\d+[,]\d{2}))
   last_line: ^\s+(?P<product>(Serienummer[:]\s\w+(?:\s\S+)*))
   line: ^\s+(?P<product>(\w+(?:\s\S+)*))\s+(?P<qty>[-]?\d+)\s*$
 - start: 'Artikel'
   end: 'Exclusief BTW'
   line: ^\s+(?P<product>(\w+(?:[:|.]|\s\S+)*))\s+(?P<qty>[-]?\d+)\s+(?P<tax_percent>\d+)[%]?\s+[€]?\s(?P<price_subtotal>([-]?\d{0,3}?[.]?\d+[,]\d{2}))
   types:
     qty: float
     price_unit: float
     price_subtotal: float
     tax_percent: float
```

Yet, the current code cannot support this kind of syntax. 
Converted it to the current syntax. But had to comment out the second line rule. 
```
fields:
  lines:
     parser: lines
     start: 'Artikel'
     end: 'Exclusief BTW'
     first_line: ^(?P<product>(\S+(?:\s\S+)*))\s+(?P<qty>[-]?\d+)\s+[€]?\s+(?P<price_unit>(\d{0,3}[.]?\d{0,3}[,]\d{2}))\s+(?P<tax_percent>\d+)[%]?\s+[€]?\s(?P<price_subtotal>([-]?\d{0,3}[.]?\d+[,]\d{2}))
     last_line: ^\s+(?P<product>(Serienummer[:]\s\w+(?:\s\S+)*)) # \s*$
     line:
       - ^\s+(?P<product>(\w+(?:\s\S+)*))\s+(?P<qty>[-]?\d+)\s*$
       # next line is for parsing thuiskopie heffing
[...]

PR #407, opened 2022-09-25
parsers: lines: support "rules" for multiple sets of regexes

rmilecki:
```
templates: make Amazon YAML use "rules" for "lines"

This doesn't really change anything and isn't necessary. It however
allows testing "rules" implementation in the "lines" parser.
```
```
lines: support "rules": field for multiple sets of parsing regexes

Sometimes companies use more than 1 format for line-parseable data. They
may e.g. randomly add some extra columns that are used occasionally.

This commit adds "rules" field support to the "lines" parser. It allows
defining multiple sets or regexes ("start", "end", "line" & friends) for
a single field.

Usage of "rules" is optional. Backward compatibility wiht existing
templates is preserved.
```

PR #463, opened 2023-02-04
lines: support "rules" field for multiple sets of parsing regexes

rmilecki:
```
Sometimes companies use more than 1 format for line-parseable data. They
may randomly generate invoices with e.g.
1. Some extra columns that are used occasionally
2. Rearrange columns order

Such format changes may be too invasive to support parsing with e.g.
multiple "line" regexes.

This commit adds "rules" field support to the "lines" parser. It allows
defining multiple sets or regexes ("start", "end", "line" & friends) for
a single upper field.

Usage of "rules" is optional. Backward compatibility wiht existing
templates is preserved.
```

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -137,6 +137,10 @@ This parser allows parsing selected invoice section as a set of lines
 sharing some pattern. Those can be e.g. invoice items (good or services)
 or VAT rates.
 
+Some companies may use multiple formats for their line-based data. In
+such cases multiple sets of parsing regexes can be added to the `rules`.
+Results from multiple `rules` get merged into a single array.
+
 It replaces `lines` plugin and should be preferred over it. It allows
 reusing in multiple `fields`.
 
@@ -148,6 +152,17 @@ Example for `fields`:
         start: Item\s+Discount\s+Price$
         end: \s+Total
         line: (?P<description>.+)\s+(?P<discount>\d+.\d+)\s+(?P<price>\d+\d+)
+
+    fields:
+      lines:
+        parser: lines
+        rules:
+          - start: Item\s+Discount\s+Price$
+            end: \s+Total
+            line: (?P<description>.+)\s+(?P<discount>\d+.\d+)\s+(?P<price>\d+\d+)
+          - start: Item\s+Price$
+            end: \s+Total
+            line: (?P<description>.+)\s+(?P<price>\d+\d+)
 
 ### Legacy regexes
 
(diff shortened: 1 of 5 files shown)


rmilecki:
This is a refactored version of https://github.com/invoice-x/invoice2data/pull/407. Rebased, with updated description and test case added.

I think that added test provides a good idea why we may need this feature. The thing is it's not always possible to parse lines of varying columns with multuple regexes. Look at the example for included test case:
```
No  Angle [°]  Length [cm]
1   30         30
2   45         40
3   90         60
Count: 3

No  Length [cm]  Angle [°]
4   80           135
5   100          180
Count: 2
```
If some company changes orders of columns, we need different set of rules for parsing such lines. It's bacause it's impossible to write a generic `line` regex that will recognize whether the number is "Angle" or "Length".

Let me know what do you think about such feature.

bosd:
Thanks for this PR. It does what it is supposed to do. :+1: :tada: 

Maybe it was not the intention of this PR.
Yet, I was hoping it to do more and to provide a solution for the problem as described here:
https://github.com/invoice-x/invoice2data/pull/407#issuecomment-1257176908


_The possibility to parse multiple sets of regexes while applying the extraction rules (fist_line, last_line) and retaining the order of the input string in the output._

**Technical Background behind previous pr 378**
In the previous PR I achieved it by nested for loops. :skull:  :warning: (yes maybe there are better iteration methods)
The main loop was going trough each line of optimized_str. (plain and simple)

Then there was the nested for loop applied on that particular line.
 The nested for loop, was going over the sets of regexes (aka rules). Trying to find a match. Yet also taking into account if the regex match was between the first_line and last_line of that particular rule. 
 If all the conditions are met, it was appended to the output.
 
 (If it was not met detailed feedback was provided. Line: XXX matched, but is not between firstline and lastline, so ignoring)


The fundamental difference seems to be that the code in **this** pr is considering the order of rules how they where written in the template file. (assumption). Instead of considering the order from the optimized string from the input file.

------------------------------------------------

What to do with this pr?? We can merge it?? 
However, I still hope to get the initial problem solved. 
I hope the explanation in this comment was helpfull how to attack the problem.
(only now implement it with better code then nesting multiple loops)

rmilecki:
@bosd: I believe we can parse Mekro invoices the way you expected since #417.

I prepared this pull request to handle different cases.

Please kindly take another look at #417. I think you managed to parse Mekro invoices by using two regex for the `lines`.

bosd:
Will later look at the mekro example again.

Tested this pr, with an template generated to solve #428.
Works, perfectly! Thanks @rmilecki

bosd (approved):
Functional Tests :+1:

Issue #444, opened 2023-01-08
remove_accents option does not work properly anymore


alainv62:
Since the removal of the unidecode library and its replacement with the unicodedata module in commit https://github.com/invoice-x/invoice2data/commit/4d517d131fdaa568b4282f2f0a12abef98751075, the remove_accents option does not work properly anymore.
Eg: in French, 'référence' is replaced with 'rfrence'.
It seems that the normal form KC is here responsible as the normal form KD works fine with this example ('référence' is properly replaced with 'reference').

esmeraldas63:
Have encountered same issues with Lithuanian and Latvian letters too

bosd:
We need to fix this! Currently im travelling. So not much time to attend to this.
Does anyone have a suggestion for an alternative implementation or library?

alainv62:
Why not just revert to the former solution `optimized_str = unidecode(optimized_str)` which was working fine?

bosd:
> Why not just revert to the former solution ?

Because of the mentioned license issue.

There are some good alternatives. Like built-ins or alternative libraries ( adding extra dependencies is not preferred).

alainv62:
What about using the normal form KD ?

rmilecki:
> Why not just revert to the former solution `optimized_str = unidecode(optimized_str)` which was working fine?

See description in pull request that introduced that change: https://github.com/invoice-x/invoice2data/pull/436

And for the orginal report: https://github.com/invoice-x/invoice2data/issues/435

bosd:
> What about using the normal form KD ?

Yes, that seems to fix it..
small test:
```
>>> str1="référence"
>>> unicodedata.normalize('NFKD', str1).encode('ascii', 'ignore').decode('ascii')
'reference'
```


```
>>> str2="ä"
>>> unicodedata.normalize('NFKD', str2).encode('ascii', 'ignore').decode('ascii')
'a'
```

bosd:
It is probably an good idea to add a test for the remove accents function.
Is there an example invoice to add to the tests?

PR #464, opened 2023-02-05
remove_accents: refactor to NFKD normalization

bosd:
Refactor the remove accents functions to NFKD normalization form. So référence get normalized to reference.

This fixes: #444

### DIFF
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -82,7 +82,7 @@ class InvoiceTemplate(OrderedDict):
 
         # Remove accents
         if self.options["remove_accents"]:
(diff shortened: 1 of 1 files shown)


rmilecki:
Looks good, could we also add some test for that?

bosd:
> Looks good, could we also add some test for that?

I agree that it is advisable to cover it in tests.

My preference is to fasttrack this one. Generate a new release on pypi. As there are some minor bugs in the current release.

Then in a separe PR create more advanced tests for the normalize function.
(I am propably not the one to write these advanced tests because we don't have it in my language.)
I could imagine someone wants to add test for german umlaut (was mentioned somewhere here)
As well as the polish & french language might hase some characters which might be affected by this..

@rmilecki If you are ok, with fasttracking this one in the current form w/o tests. Please merge this one 👍

bosd:
Sadly, This is not a correct solution.  It now also deletes the € sign.

PR #468, opened 2023-02-10
tesseract: [FIX] pdf pre-processing

bosd:
Before this PR the mimetype of the input file was detected by it's extension.
It works fine for cli usage.

It fails when invoice2data is used as a library and a tempfile/stream is used as an input. (as there is no extension available)
Making it impossible to parse pdf files when using invoice2data  as a library. 
As tesseract cannot handle pdf files as an input. 

After this pr, the mimetype is also detected when using invoice2data as a library.

@m3nu , @rmilecki I'm consideirng this as a hotfix, so fasttracking this one.

### DIFF
--- a/src/invoice2data/input/tesseract.py
+++ b/src/invoice2data/input/tesseract.py
@@ -2,6 +2,7 @@
 
 from distutils import spawn
 import tempfile
+import mimetypes
 
 from subprocess import Popen, PIPE, STDOUT, CalledProcessError, TimeoutExpired
 from subprocess import run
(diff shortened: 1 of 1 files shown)


rmilecki:
Thanks for handling this.

bosd:
It turns out, this really did not make a difference. The guess_type function also looks at the file extension.
Instead of running low level compares on the content.

It turned out that the problem was in the program calling invoice2data.
It was passing a tempfile without extension.
Forcing the program to always pass the `.pdf` extension fixed it.

PR #469, opened 2023-02-10
replace distutils with shutil

bosd:
The package distutils is deprecated and slated for removal in Python 3.12. We are using the function find_executable that belongs to that package. We should instead be using which from the package shutil.


rmilecki (approved):
Tests are passing, nothing to object.

PR #470, opened 2023-02-12
Performance update: match keywords on extracted_str

bosd:
Before this PR for each individual template an optimized string was generated.
This impacts the performance negatively, specifically if one has a lot of templates.

This PR greatly increases performance as an optimized_str is only generated on the matched template instead of on all templates.

On my local system, I realized a 2x performance increase.

:warning: Warning: Every performance increase comes at a cost. It might break some templates.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -381,9 +381,7 @@ Look at the invoice and find the best identifying string. Tax number +
 company name are good options. Remember, *all* keywords need to be found
 for the template to be used.
 
-Keywords are compared *after* processing the extracted text. So if you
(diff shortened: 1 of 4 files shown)


rmilecki (commented):
I think it's acceptable. If some invoices require more complex matching, one can always use a regex in the `keywords` (just like in `com.flipkart.WSRetail.yml` as modified in this pull request).

@m3nu: do you have an opinion on this?

rmilecki:
Ping @m3nu

m3nu:
Should be OK to match an invoice *before* doing the optimizations for extractions. Maybe add a graph of the steps at some point to make it quicker to understand and start?

PR #475, opened 2023-02-14
Some UX Improvements, Colored formatting debugging

bosd:
These small improvements might make it easier to debug your templates.

- Colors on the matching / regex lines
image

- Improved error messages
- Fix indentation for lines (Making them easier to read)
- Table plugin, see which ruleset is tested
image

- Formatted json style output while debugging, so it is easier to recognize the different lines.
image

You can see it in action in the github actions check of this pr.


bosd:
Some rationale about moving back to `%s` string format:

Usage of f strings in the logger are [8% slower](https://rmsol.de/2019/09/08/Logging-Benchmark/) slower then %s format.

Issue #467, opened 2023-02-06
remove_accents not covered by unit tests


bosd:
As discovered in #464 the remove_accents function is not covered by any tests.

Users have reported problems with umlauts, accents and some other characters.
Would be nice to have an advanced unit test for this.

bosd:
We need to add a test, that currency signs does not get replaced.
As mentioned in:
#477 remove_accents option deletes € sign

bosd:
fixed in #494

Issue #477, opened 2023-02-18
remove_accents option deletes € sign


bosd:
since #464 the remove_accents function also deletes the € sign.

bosd:
Would this be an alternative implementation for the remove_accents?
https://stackoverflow.com/a/71408065
It also relies only on roce python libs.

PR #478, opened 2023-02-18
tests: custom: add basic cases for "options"

rmilecki:
```
This adds basic tests for "remove_accents" and "replace".
```

### DIFF
--- a/tests/custom/options-accents.json
+++ b/tests/custom/options-accents.json
@@ -1,0 +1,12 @@
+[
+    {
+        "issuer": "Options Accents Test",
+        "date": "2023-02-18",
+        "invoice_number": 1234,
+        "amount": 99.99,
+        "german": "Tschuss",
+        "polish": "Szczescie",
+        "currency": "EUR",
+        "desc": "Invoice from Options Accents Test"
+    }
+]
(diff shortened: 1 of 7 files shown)


rmilecki:
This resolves #467

It doesn't cover all accents cases. Especially test for handling € character will need to be added while working on the #477

bosd:
Thanks for your contribution. Yet, my first look at this. The solution in this pr looks complex.

On a second thought, I think adding tests for the remove_accents is unnecessary and overly complicating things.
I think we should treat the breaking of the function as an anomaly.

With the new proposal (#479) for this function to remove only the combining diacritical marks in a certain range.
It is unexpected to change the strings and leave out characters.
(like with the previous ascii and unidecode method)



just to share my train of thoughts....
Earlier I was more thinking about generating and unittest for functions.
test_functions.py

```
test_remove_accents()
string_to_test  = "é€$%^&*@!.a Málaga François Phút Hơn 中文ß"
expected_result = "e€$%^&*@!.a Malaga Francois Phut Hon 中文ß"
result = remove_accents(string_to_test)
assertIsEqual( expected_result, result)
```

Note: for that to work, the remove_accents needs to get it's own function. Increasing function calls again.
I think this is all unnecessary and unwanted. Just keep it simple :smiley:

rmilecki:
Obsoleted by a better implementation in the #494

PR #479, opened 2023-02-19
[FIX] remove_accents

### LINKED ISSUE #477: remove_accents option deletes € sign

bosd:
Use an regex method to remove the combining diacritical marks

https://en.wikipedia.org/wiki/Combining_Diacritical_Marks

fixes #477

### DIFF
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -82,7 +82,7 @@ class InvoiceTemplate(OrderedDict):
 
         # Remove accents
         if self.options["remove_accents"]:
-            optimized_str = unicodedata.normalize('NFKD', optimized_str).encode('ascii', 'ignore').decode('ascii')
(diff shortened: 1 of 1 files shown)


bosd:
Some context:
To come up with this solution I tested a handfull of different methods to solve this.
Eventually I picked the one which did deliver the desired outcome and best performance.

I'm to humble to share the messy code of these tests. But here is the performance graph.

remove_accents_perf

strip_accent function was clearly the slowest performing, when the input string increases.
So eventually I went for the one with the `re` one (shown in red).
The orange plot is the current implementation of remove accents. It's the fastest. But removes too much characters.
So not an option.

The purple plot is the same function as the red one. Only the python `.re` module has been exchanged for the `regex` module.

rmilecki (approved):
Quickly tested with my WIP tests coverage in #478, seems to work.

bosd:
Should we hotfix this? by making another release on pypi.

Issue #392, opened 2022-09-08
Return extracted_str if no templates found with extract_data()  (python) ?


Whaoo:
Hi, 

Is there a way to return the extracted_str (full pdf text in str) if no templates are found for the pdf ? 

Saw it in the main.py that in debug extracted_str is exactly what i want to collect, that would save me time rather than calling and storing again pdf2text.

Is there any way to return it in extract_data() if no templates are found for the .pdf ? 

Many thanks

bosd:
Is this what you are looking for? Or get some inspiration from?
Did'nt test this.

https://github.com/OCA/edi/pull/399/files#diff-652ac3ae132c668bf2ac61903174bbc0c254c98bf549aac7cad47a515259ed32R70-R128

rmilecki:
Maybe we could make `invoice2data` more object oriented?

```
# Use static method
templates = Invoice2Data.read_templates("templates/")

i2d = Invoice2Data()
try:
    i2d.extract_data("foo.pdf", templates=templates)
except Exception as e:
    print('Failed to extract data: ' + str(e))
    print('Extracted text: ' + i2d.get_extracted_text())
```

legalsylvain:
Hi @rmilecki
I'm looking for a way to have the detail of the parsing error. (no templates found / missing required feld / ...).  For the time being, the information is in the log, but not accessible if using invoice2data as a library.

what I don't understand in your code, is that AFAIK, extract_data doesn't raise an error. Or did I missed something ?
```
try:
    i2d.extract_data("foo.pdf", templates=templates)
except Exception as e:
    print('Failed to extract data: ' + str(e))
    print('Extracted text: ' + i2d.get_extracted_text())
```

Whaoo:
Hi guys, nice to see my question is interesting other people 
Will try using what you pushed @rmilecki :)

PR #480, opened 2023-02-19
Provide Invoice2Data class for object oriented usage

rmilecki:
```
This will allow using invoice2data in an object oriented way like:

import invoice2data

i2d = invoice2data.Invoice2Data()
i2d.extract_data("foo.pdf")
```

### DIFF
--- a/src/invoice2data/__init__.py
+++ b/src/invoice2data/__init__.py
@@ -1,1 +1,2 @@
 from .main import extract_data  # noqa: F401
+from .main import Invoice2Data  # noqa: F401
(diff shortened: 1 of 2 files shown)


rmilecki:
I was wondering if there is any interest in providing an `Invoice2Data` class so this project can be used as a library in an object oriented way.

That idea originally came from #392 but maybe it could be a generic improvement.

m3nu:
Sounds good. So the class would be instantiated with some templates and then read invoices? Makes sense, since we already keep some state via templates and the input source.

m3nu:
I'd still carefully consider which objects make sense. Maybe you also want an `Invoice` class to hold the read invoice.

bosd (approved):
Good idea! :+1

bosd:
Re-implemented on current master as 8f99b55 (rebased past the main.py→__main__.py refactor; added type hints + a test). `Invoice2Data` is now exported from the package root. Thanks @rmilecki!

PR #481, opened 2023-02-25
pdplumber return empty string on importerror

### LINKED ISSUE #362: Template validator ?
Hello all,
i would like to run i2d with something like --check-template mytemplate.yml just for 

1. yml syntax validation
2. grammar validation (fields ok / nok, i2d version compatibility)
3. regexp validity (for example if you put a "/" or something like regex parser can't understand)
4. missing fields warning (for example "there is no amount regex in this yml"

That sort of tool can really help us making good yml files ... and "debug" yml files

bosd:
Was running some tests, encountered following error when pdfplumber is not available.
This PR returns and empty value and let invoice2data fail gracefully.

Before:
`invoice2data input.pdf --input-reader=pdfplumber`

```
DEBUG:invoice2data.input.pdfplumber: Cannot import pdfplumber
  File "/home/emiel/.local/bin/invoice2data", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/emiel/.local/lib/python3.11/site-packages/invoice2data/main.py", line 312, in main
    res = extract_data(f.name, templates=templates, input_module=input_module)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/emiel/.local/lib/python3.11/site-packages/invoice2data/main.py", line 166, in extract_data
    extracted_str = input_module.to_text(invoicefile).decode("utf-8")
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/emiel/.local/lib/python3.11/site-packages/invoice2data/input/pdfplumber.py", line 29, in to_text
    with pdfplumber.open(path, laparams={"detect_vertical": True}) as pdf:
         ^^^^^^^^^^
UnboundLocalError: cannot access local variable 'pdfplumber' where it is not associated with a value

```

After:

```
ERROR:invoice2data.input.pdfplumber: Cannot import pdfplumber
ERROR:root: Failed to extract text from testin.pdf using invoice2data.input.pdfplumber
```

fixes #362


src/invoice2data/input/pdfplumber.py
```diff
@@ -19,7 +19,8 @@ def to_text(path):
     try:
         import pdfplumber
     except ImportError:
-        logger.debug("Cannot import pdfplumber")
+        logger.error("Cannot import pdfplumber")
+        return "".encode("UTF-8")
```
rmilecki, line 23: Returning empty string suggests that invoice was parsed but was empty.
If we want to return some value then make it `None` please.

If you take a look at `pdftotext.py` however, you'll see it raises `EnvironmentError` if `pdftotext` is missing. So returning `None` will make `pdfplumber.py` somehow incompatible with the `pdftotext.py`.

As for decision what is better: `return None` or `raise EnvironmentError` - I have no idea or preference.

rmilecki (changes requested):

src/invoice2data/input/pdfplumber.py
bosd: @rmilecki I agree, we need to think of an solution for this.

Returning None conflicts with:
https://github.com/invoice-x/invoice2data/blob/a5bdd50b1cfc1e7953b0d1467c9915bafacbc3aa/src/invoice2data/main.py#L89

as Nonetype cannot be decoded
(Maybe we can remove the decode line? I assume it is a python2 leftover)

pdftotext might be a different story, as it is one of the default/main parsers.
So making everything fail when it's unavailable is not a big dead.

Is there a way to raise the error, only if the pdfplumber input module is called?
We don't want the whole lib to fail on this missing requirement.

This pr https://github.com/invoice-x/invoice2data/pull/491
is currently failing because of the missing pdfplumber.
(the test should not even run when it is unavailable but that's a different sunbject)
In that example, there should be an `ImportError` or `EnvironmentError`

bosd:
Re-implemented on current master as 8f99b55: pdfplumber.to_text() now raises ImportError cleanly (it previously fell through to a NameError), and extract_data() returns {} if an input module raises. Thanks!

PR #483, opened 2023-02-27
Unify Template tests & add Loader tests

bosd:
Some cherry picks of https://github.com/invoice-x/invoice2data/pull/457
Thanks @timorieber


bosd:
Oops, latest commit depends on #471

PR #487, opened 2023-03-03
Regex: String joining in case of multiple matches

bosd:


bosd:
Merging this one now, according to the "EAFP".
So I can continue working on the template PR, this is blocking the tests.

Propably later will, merge more pr's following th EAFP.

easier to ask forgiveness than permission.

PR #489, opened 2023-03-08
line parser more verbose debugging

bosd:
Make the debugging logger more in line with the regex and tables parser.

### DIFF
--- a/src/invoice2data/extract/parsers/lines.py
+++ b/src/invoice2data/extract/parsers/lines.py
@@ -25,6 +25,10 @@ def parse_block(template, field, settings, content):
     # Validate settings
     assert "line" in settings, "Line regex missing"
 
+    logger.debug(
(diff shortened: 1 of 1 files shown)


src/invoice2data/extract/parsers/lines.py
```diff
@@ -25,6 +25,10 @@ def parse_block(template, field, settings, content):
     # Validate settings
     assert "line" in settings, "Line regex missing"
 
+    logger.debug(
+        "START line block content ========================\n%s", content
+    )
+    logger.debug("END line block content ==========================")
```
rmilecki, line 31: Can you replace "line" with "lines", please?

bosd: :+1 Good catch!

rmilecki (approved):

PR #490, opened 2023-03-11
Improve Template Error Messages

bosd:
Using invoice2data as I lib.
Before I got this error message in the upstream application.
image

In the frontend the logger is not visible.
It is more usefull to show which template contains the error.

### DIFF
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -63,7 +63,10 @@ class InvoiceTemplate(OrderedDict):
             self.options.update(self["options"])
 
         for lang in self.options["languages"]:
-            assert len(lang) == 2, "lang code must have 2 letters"
+            assert len(lang) == 2, (
+                "Error in Template %s lang code must have 2 letters"
+                % self["template_name"]
+            )
 
         # Set issuer, if it doesn't exist.
         if "issuer" not in self.keys():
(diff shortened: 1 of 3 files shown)


src/invoice2data/extract/invoice_template.py
```diff
@@ -121,9 +127,10 @@ def matches_input(self, optimized_str: str) -> bool:
             return False
 
     def parse_number(self, value):
-        assert (
-            value.count(self.options["decimal_separator"]) < 2
-        ), "Decimal separator cannot be present several times"
+        assert value.count(self.options["decimal_separator"]) < 2, (
+            "Error in Template %s \nDecimal separator cannot be present several times"
```
rmilecki, line 131: The other 7 asserts don't use `\n` in messages, I'd suggest dropping `\n` from this one for compatibility.

src/invoice2data/extract/invoice_template.py
```diff
@@ -90,7 +93,10 @@ def prepare_input(self, extracted_str: str) -> str:
 
         # Specific replace
         for replace in self.options["replace"]:
-            assert len(replace) == 2, "A replace should be a list of exactly 2 elements."
+            assert len(replace) == 2, (
+                "Error in Template %s \nA replace should be a list of exactly 2 elements."
```
rmilecki, line 97: The other 7 asserts don't use `\n` in messages, I'd suggest dropping `\n` from this one for compatibility.

rmilecki (approved):

PR #491, opened 2023-03-11
raise error on missing required fields

bosd:
Raise an value error on missing required fields,

https://github.com/invoice-x/invoice2data/issues/392#issuecomment-1439183201


bosd:
depends upon: https://github.com/invoice-x/invoice2data/pull/492

PR #493, opened 2023-03-12
Refactor to_text() to return string instead of bytes

rmilecki:
```
Function name "to_text" suggests it should return a text. Python uses
str type for storing texts. Having that function return bytes was
counter-intuitive.

This also simplifies code as most input methods deal with str type. With
this change there is no need to encode str into bytes and decode it
back.
There are actually only 2 input methods dealing with bytes: "pdftotext"
and "tesseract". Make them decode bytes into str before returning from
to_text().
```


rmilecki:
@bosd: I thought about this small refactoring after your comment in the #481. Do you think it's OK?

bosd (approved):
Makes sense!! :+1:

PR #498, opened 2023-03-18
Support coerce types in table plugin

bosd:


bosd:
Quickly drafted this. Did not test it yet. Seems good tests are green :smile:

PR #499, opened 2023-03-18
Error message on calling unknown plugin

bosd:
Accidently discovered this was missing on the template validation.
tried to call
`tables:`
but writing
`table:`
in the template didn't produce an error

Related to #362


bosd:
code needs refactoring to work.

PR #509, opened 2023-04-24
Avoid UnboundLocalError when exception is raised

mcweidner:
Adding `continue` to the except block, otherwise an `UnboundLocalError: local variable 'tpl' referenced before assignment` is thrown when trying to assign `tpl["template_name"] = name` after `json.loads()` raised an exception.

@bosd I think reading the templates should not fail completely because of one invalid template. Maybe we could also be more sensitive about the file types and only try `json.loads` for `.json` files instead of every file except .yml files (and for other file types raise a warning that only .yml and .json are supported)? What do you think about this?


bosd:
> Maybe we could also be more sensitive about the file types and only try json.loads for .json files instead of every file except .yml files

Good Idea, I was unaware of the fact folders could also contain different files.

bosd (approved):

PR #520, opened 2023-08-06
loader: require .json extension and allow .yaml extension

rmilecki:
```
loader: allow YAML templates with .yaml extension

Oficially suggested extension for YAML files is .yaml. Accept it.
```
```
loader: require .json extension for JSON files

Trying to parse every file other than *.yml as JSON file isn't the best
idea. It wastes time on trying to parse non-template files. It produces
warnings like "json Loader Failed to load (...)template" for files that
were never meant to be templates. It's a bad pracice in general.

Check for file extension before using json.loads().
```

### DIFF
--- a/src/invoice2data/extract/loader.py
+++ b/src/invoice2data/extract/loader.py
@@ -72,12 +72,14 @@ def read_templates(folder=None):
                     except YAMLError as error:
                         logger.warning("Failed to load %s template:\n%s", name, error)
                         continue
-                else:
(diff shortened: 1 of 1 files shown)


bosd:
Thanks for this PR.
I agree the loader could use some work. It's on my wish list :)
As the recent refactoring broke the possibility to stream load templates.

I think adding more extension filters will block the possibility to streamload templates even further.

PR #522, opened 2023-08-12
Clean up some build dependencies

tjni:
While `pip` and `pytest-runner` could be development dependencies, I don't think they are needed for the actual build by setuptools. I'm hoping to clean this up to simplify in a small way packaging this in [nixpkgs](https://github.com/NixOS/nixpkgs).


bosd (approved):
LGTM

PR #524, opened 2023-09-02
TUTORIAL: Add "vatin" to standard fields

rmilecki:
```
In most countries end customers pay a VAT (value-added tax). Companies
issuing invoices with VAT need to have a unique identification number
assigned. Local names and abbreviations of those numbers differ across
countries (UID, DIČ, CVR, CBL, NIP, NIF, etc.).

European Union seems to reference those numbers as "VAT identification
numbers". Some open projects like Wikipedia and OpenStreetMap use
"vatin" abbreviation for it. Add field of the same name to the TUTORIAL.
```

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -79,6 +79,7 @@ Standard fields:
 - `date`: the date invoice was issued
 - `invoice_number`: unique number assigned to invoice by an issuer
 - `amount`: total amount (with taxes)
+- `vatin`: [VAT identification number](https://en.wikipedia.org/wiki/VAT_identification_number)
 
 ### Parser `regex`
 
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -79,7 +79,7 @@ Standard fields:
 - `date`: the date invoice was issued
 - `invoice_number`: unique number assigned to invoice by an issuer
 - `amount`: total amount (with taxes)
-- `vatin`: [VAT identification number](https://en.wikipedia.org/wiki/VAT_identification_number)
+- `vat`: [VAT identification number](https://en.wikipedia.org/wiki/VAT_identification_number)
 
 ### Parser `regex`
 


rmilecki:
* https://en.wikipedia.org/wiki/Value-added_tax
* https://wiki.openstreetmap.org/wiki/Key:ref:vatin

bosd:
Thanks for this PR.
Yet, currently the name `vat` is in use for this purpose.
At least in the templates contributed by me and alexis-via.

We use this number as the first method of matching the parsed invoice with the partner in the accounting software.

It appears this is one of the undocumented items of this module.
Is it ok to use the name `vat`?

bosd:
On a invoice there are generally 2 VAT numbers as required by law of some EU countries.
One is the vat number of the issuer, the other is of the receipient.

For the recipient vat number we use:

 fieldname | type | Description |
| -------------- | :---------: | :-------------------------------------- |
 company_vat | char | The vat number of the company to which the invoice is addressed to. Used to check if the invoice is actually is adressed to the company which wants to process it. (Very useful in multi-company setup)  |

TUTORIAL.md
```diff
@@ -79,6 +79,7 @@ Standard fields:
 - `date`: the date invoice was issued
 - `invoice_number`: unique number assigned to invoice by an issuer
 - `amount`: total amount (with taxes)
+- `vatin`: [VAT identification number](https://en.wikipedia.org/wiki/VAT_identification_number)
```
bosd, line 82:
```suggestion
- `vat`: [VAT identification number](https://en.wikipedia.org/wiki/VAT_identification_number)
```

bosd (changes requested):

bosd:
Ping @rmilecki

bosd (approved):
Changed the field to `vat` as used in majority of templates.

Issue #505, opened 2023-04-03
Value that is zero is matched but then not parsed


coding4me2:
If I have a field in which the the value is zero, such as VAT @ 0% which yields zero, then it will not return the field as either type int or float. That's using either legacy usage of amount_name_of_field or using parser regex and setting the type to float or int. As you can see it does find the '0.00' string but then says it failed to parse it.

`DEBUG:invoice2data.extract.parsers.regex:field=amount_total_vat | regex=Total Amount \(GBP\)\s+\d+.\d+\s+(\d+.\d{2}) | **matches=['0.00']**
ERROR:invoice2data.extract.invoice_template:Failed to parse field amount_total_vat with parser regex
`

With zero rated and exempt VAT at 0% it can be fairly common to get a vat amount of 0.00 on an invoice.

PR #538, opened 2023-11-24
Support zero values also in legacy syntax

dannyadair:
https://github.com/invoice-x/invoice2data/pull/507 fixed the issue with zero values reported in https://github.com/invoice-x/invoice2data/issues/505 but not for legacy syntax.


bosd:
Nice spot! Thanks for the fix 👍

bosd (approved):
Looks Good To Me! 👍

PR #555, opened 2024-04-25
Draft:Windows specific prefix check (magick), add support for windows in github actions

alexm96:
Noticed an issue on windows that just using convert isn't correct as it conflicts with a native command in windows. Instead just checking if magick exists should be enough, then convert can be prefixed with that when running the full command. Also ran black for tesseract.py, which unfortunately reformatted some stuff :(


bosd:
Good catch. Thanks for this PR.
Is it possible to separate the black formatting in a separate commit?

I just realized that unit tests here on gh, don't test it on windows.
Can you add those to gh workflow? 🙏

alexm96:
@bosd sure thing, I'll switch it to draft for now and play around with it :)

alexm96:
@bosd btw, i've modified the main.yml accordingly. Have been playing around with it on my forked branch and some tests are failing, so i'll keep grinding away at it. I'll probably end up splitting it out , since there seems to be some issues with windows GA running in parallel, where they fail to lock. Anywaz, i'll keep you posted in here once it's ready to review :)

bosd:
Thanks for the update.
Earlier, I made attempt to let the actions run here.
I'll hold off for now, as you can test it on your own repo.

bosd:
Maybe this is of help. Tests are running here correctly on a matrix of ubuntu releases and one windows.
https://github.com/py-pdf/pypdf_table_extraction/blob/main/.github/workflows/tests.yml

alexm96:
Yeah, if it's fine to only test against one Python version I would do it like that. Otherwise I would run the windows ones sequentially :)

bosd:
Oopsie, It looks like I messed up this PR, while trying to resolve merge conflicts

bosd:
superseeded by #565
@alexm96 Can you help fixing the tests. 

>Yeah, if it's fine to only test against one Python version I would do it like that. 

Sounds like an good idea, Windows tests take a lot of compute and time to complete when running against the full matrix.

bosd:
Superseded by #565 (merged 2026-07-20 as `5f09f5e0`), which landed the same fix: use ImageMagick 7's `magick` command when available (with fallback to legacy `convert` on IM6 systems) so the pre-conversion pipeline works on Windows where the built-in `convert.exe` shadows it. Thanks @alexm96 for the original spot!

AlexOGTv:
@bosd thanks for the update! Happy it was eventually fixed :)

PR #565, opened 2024-07-19
add support for windows, add windows specific prefix command (magick), and github actions

bosd:
superseding #555


bosd:
On windows there is trouble finding the correct tesseract command.
This needs fixing, maybe something like done here:
https://github.com/ocrmypdf/OCRmyPDF/blob/main/src/ocrmypdf/subprocess/_windows.py

PR #642, opened 2025-08-13
Fix mypy typing

bosd:


bosd:
Closing as superseded by #692 (PEP 585/604 type-hint modernization), which rewrote the annotations across the codebase. The remaining type-checking gap will be handled when we adopt `ty` with the new template. Thanks!

PR #652, opened 2025-09-19
Add number extraction method and line skipping functionality

avonbuttlar:
- Implemented `extract_number_from_text` to extract numeric values from strings with text.
- Updated `coerce_type` to utilize the new extraction method for converting strings to int/float.
- Added support for skipping lines based on provided patterns in the `parse_block` function.

Since i dont get the tests to run (i do not have tesseract), i hope someone can help :).

Since this is my first pr to a public repo ever, please give feedback. Ik i need to do some documentation still, but i tought i first add the functionalltiy.

### DIFF
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -204,6 +204,27 @@ class InvoiceTemplate(OrderedDictType[str, Any]):
         logger.debug("result of date parsing=%s", res)
         return res
 
+    def extract_number_from_text(self, value: str) -> str:
+        """Extracts the first numeric value from a string containing text.
+
+        This function finds and extracts numeric values from strings that may
+        contain text like "4 Stück", "12123 Stk.", "€25.50", etc.
+
+        Args:
+            value (str): The string containing numbers and possibly text.
+
+        Returns:
+            str: The extracted numeric value as a string, or empty string if no number found.
+        """
+        # Pattern to match numbers with optional decimal separators
+        # This matches: 123, 123.45, 123,45, 1.234,56, 1,234.56, etc.
+        pattern = r'[-+]?\d{1,3}(?:[.,\s\']\d{3})*(?:[.,]\d+)?'
+        
+        match = re.search(pattern, value)
+        if match:
+            return match.group().strip()
+        return ""
+
     def coerce_type(self, value: str, target_type: str) -> Any:
         """Coerces a value to the specified target type.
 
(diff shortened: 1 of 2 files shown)


bosd:
Thanks for the PR @avonbuttlar, and welcome — nice first contribution! 🎉

The `skip_line` half is a useful, well-contained addition. The `extract_number_from_text` half needs rework before it can be merged, though, because `coerce_type` applies it to **every** `int`/`float` field globally, and the current regex truncates numbers ≥ 1000 that do not use a thousands separator.

The pattern `[-+]?\d{1,3}(?:[.,\s']\d{3})*(?:[.,]\d+)?` assumes thousands-grouping, so:

| input | extracted | expected |
|---|---|---|
| `1234` | `123` | 1234 |
| `1234.56` | `123` | 1234.56 |
| `1939,00` | `193` | 1939 |
| `12123 Stk.` | `121` | 12123 |

So even the `12123 Stk.` example from the description comes out as `121`. Applied to all numeric fields, this would change/break extraction of thousands-range amounts across existing templates (and the comparison fixtures in `tests/compare/`).

Suggestions:
- Make the number extraction **opt-in** (a template/field option) instead of changing the global `coerce_type` path, so existing templates are unaffected.
- Fix the regex so it does not require thousands separators (the `\d{1,3}(...)*` grouping is the culprit), and add unit tests covering `1234`, `1234.56`, `1.234,56`, `12123 Stk.`, `€25.50`, and negatives.
- Re: running tests without tesseract — that test is now skipped automatically when the `tesseract` binary is absent (latest `master`), so `pytest` should run for you after a rebase.

This is feature-sized, so it would land in a future minor release rather than the current maintenance line. Happy to help get it over the finish line — thanks again!

bosd:
Hi @avonbuttlar — picked this up and landed a cleaned-up version of both halves on the `next` branch in [fd27734](https://github.com/invoice-x/invoice2data/commit/fd27734). Short summary:

**`skip_line`** is in essentially as you wrote it (single pattern or list, dropped before the line/first_line matchers see it). 👍

**Number extraction** is reworked as an **opt-in per-field** option `extract_number: true` instead of changing the global `coerce_type` path. The regex is also corrected — the original `\d{1,3}(?:[.,\s\']\d{3})*(?:[.,]\d+)?` required thousands grouping and truncated unseparated numbers (`1234` → `123`, `12123 Stk.` → `121`), which would have silently broken every template capturing a thousands-range amount without a separator. The new regex `[-+]?\d+(?:[.,\s\']\d+)*` keeps the full number (`1234`, `1.234,56`, `1,234.56`, `25.50`, `12123` all match correctly).

15 regression tests in `tests/test_issue_652.py` lock in:
- `1234` → `1234` (no truncation)
- `1234.56` → `1234.56`
- `1.234,56` → `1.234,56`, `1,234.56` → `1,234.56`
- `12123 Stk.` → `12123`, `€25.50` → `25.50`, `-42` → `-42`
- Off-by-default semantics so existing templates are unaffected.
- `skip_line` with a string and a list of patterns.

Docs: a short reference section for both options in `docs/recommended-template-fields.md`.

Closing this PR in favor of the merged work. Thanks for the contribution — the `skip_line` half in particular slots in cleanly, and the number-extraction use case is a good addition (just needed the opt-in framing). Looking forward to more contributions from you! 🎉

bosd:
Reworked + merged on `next` in fd27734 (see comment above). Thanks @avonbuttlar — both halves of your contribution shipped.

Issue #651, opened 2025-09-18
Fix pkg_resources warning about deprecation


MonkzCode:
Hello!
I'm using invoice2data latest release version from tag.
But now I see warning:
```
python3.13/site-packages/invoice2data/extract/loader.py:14: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
```
Maybe @bosd or another maintainers can fix this one? Or maybe we should receive new release (already 2 years pass from latest release)

Thanks in advance!

MonkzCode:
@bosd , @m3nu, @alexis-via  sorry for disturbing, but I want to know are You planing fixes to this project? Or maybe new release since almost 2 years already from last one?

nomandera:
Is this fully broken now?

```
ModuleNotFoundError: No module named 'pkg_resources'
```

Any work around to get invoice2data operating again?

bosd:
@nomandera on it!

PR #653, opened 2025-09-25
use importlib to really use the installed template folder

mortbauer:
As it was before one had be basically in the repo source directory

### DIFF
--- a/src/invoice2data/extract/loader.py
+++ b/src/invoice2data/extract/loader.py
@@ -6,6 +6,7 @@ Templates are initially read from .yml or .json files and then kept as class.
 import codecs
 import json
 import os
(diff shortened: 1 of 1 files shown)


bosd:
Thanks @mortbauer — same root cause as #651, fixed in **v0.4.6** (commit `ab47786`). I kept an `os.path.dirname(__file__)`-based lookup rather than `importlib.resources.files()` because the project still supports Python 3.8 (`files()` needs 3.9+), and the installed package also needed the packaging fix (restoring `[build-system]` + package-data, which this PR didn't include) so the templates are actually bundled. Closing as resolved by the equivalent fix. https://pypi.org/project/invoice2data/0.4.6/

PR #722, opened 2026-07-29
fix(loader): skip empty template files instead of crashing (#721)

### LINKED ISSUE #721: Emtpy template results in crash
## Steps to reproduce
```sh
# Installation, etc.
mkdir demo_templates
touch demo_templates/empty_template.yaml
```

Read templates in Python:
```python
from invoice2data.extract.loader import read_templates


templates = read_templates("demo_templates")
```

## Stacktrace
```
  File "/home/philipp/proj/invoice2data/demo.py", line 4, in <module>
    templates = read_templates("demo_templates")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/philipp/proj/invoice2data/src/invoice2data/extract/loader.py", line 116, in read_templates
    tpl["template_name"] = name
    ~~~^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object does not support item assignment
```

## Images from Debugger
Image

## Possible solution
```python
if tpl is None:
    logger.warning("template %s is empty", name)
    continue
```
Image

bosd:
## Summary

- Empty `.yaml` / `null` `.json` template files no longer crash `read_templates()` with `TypeError: 'NoneType' object does not support item assignment`.
- The loader logs a warning naming the offending file and continues with the rest of the directory.

Closes #721.

## Repro (from the issue)

```py
mkdir demo_templates && touch demo_templates/empty_template.yaml
python -c "from invoice2data.extract.loader import read_templates; read_templates('demo_templates')"
# TypeError: 'NoneType' object does not support item assignment
```

`yaml.safe_load('')` returns `None`, but the next line assumes a dict and assigns `tpl["template_name"] = name`. One empty placeholder took the whole directory down.

## Test plan

- [x] `pytest tests/test_loader.py` — all 12 passing including the new `test_empty_template_is_skipped` (covers both `.yaml` and `.json` empty cases + asserts the warning is logged)

### DIFF
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -122,6 +122,23 @@ def test_template_with_missing_keywords_is_not_loaded(
     assert templates == []
 
 
+def test_empty_template_is_skipped(
+    templatedirectory: Path, caplog: pytest.LogCaptureFixture
+) -> None:
+    """Issue #721: an empty template file used to crash with
+    ``TypeError: 'NoneType' object does not support item assignment``.
+    """
+    (templatedirectory / "empty.yaml").write_text("", encoding="utf-8")
+    (templatedirectory / "null.json").write_text("null", encoding="utf-8")
+
+    with caplog.at_level("WARNING"):
+        templates = read_templates(str(templatedirectory))
+
+    assert templates == []
+    assert "Skipping empty template: empty.yaml" in caplog.text
+    assert "Skipping empty template: null.json" in caplog.text
+
+
 def test_template_name_is_yaml_filename(templatedirectory: Path) -> None:
     yamlfile = templatedirectory / "thisnameisimportant.yml"
     yamlfile.write_text(template_with_single_special_char, encoding="utf-8")
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -122,21 +122,40 @@ def test_template_with_missing_keywords_is_not_loaded(
     assert templates == []
 
 
-def test_empty_template_is_skipped(
+def test_empty_yaml_template_is_skipped(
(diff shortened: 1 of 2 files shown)


tests/test_loader.py
```diff
@@ -122,6 +122,23 @@ def test_template_with_missing_keywords_is_not_loaded(
...
 
 
+def test_empty_template_is_skipped(
+    templatedirectory: Path, caplog: pytest.LogCaptureFixture
+) -> None:
+    """Issue #721: an empty template file used to crash with
+    ``TypeError: 'NoneType' object does not support item assignment``.
+    """
+    (templatedirectory / "empty.yaml").write_text("", encoding="utf-8")
+    (templatedirectory / "null.json").write_text("null", encoding="utf-8")
```
PhilippFeO: Why do you write `"null"` into the `.json` file? Technically, this isn't an empty file anymore (and #721 is about empty templates).
I am not really familiar with the JSON spec, so I assume parsing `"null"` results in `None`. But nevertheless, why not testing against an empty `.json` as well?

bosd:
Fair point @PhilippFeO — updated in `2755121`:

- Dropped the `null.json` fixture (you're right, `null` isn't empty and #721 is specifically the empty-file case)
- Split into two tests: `test_empty_yaml_template_is_skipped` exercises the new None-guard (`yaml.safe_load('')` returns `None`), and `test_empty_json_template_is_skipped` exercises the pre-existing `ValueError` path (`json.loads('')` raises) so a future JSON-loader refactor can't quietly regress the empty-file case there either.

Thanks for the careful read!

PhilippFeO:
You're welcome!

🤝

PR #744, opened 2026-08-11
chore(docs): Decrement level of chapter 'Lines'

PhilippFeO:
Based on [Parser lines](https://invoice2data.readthedocs.io/latest/tutorial.html#parser-lines), the [Lines](https://invoice2data.readthedocs.io/latest/tutorial.html#lines) plugin is deprecated and the code also signals this to the user.

The current version of the docs implies by structure, [Lines](https://invoice2data.readthedocs.io/latest/tutorial.html#lines) is on the same level as [Parser lines](https://invoice2data.readthedocs.io/latest/tutorial.html#parser-lines) and especially [Legacy Regexes](https://invoice2data.readthedocs.io/latest/tutorial.html#legacy-regexes), that this is not deprecated. Also, [Lines](https://invoice2data.readthedocs.io/latest/tutorial.html#lines) doesn't mention it's deprecated.

I propose decrementing the level of [Lines](https://invoice2data.readthedocs.io/latest/tutorial.html#lines), such that it appears **below** [Legacy Regexes](https://invoice2data.readthedocs.io/latest/tutorial.html#legacy-regexes), implying it's deprecation automatically.

---

I ran into the issue, because I contemplated [Lines](https://invoice2data.readthedocs.io/latest/tutorial.html#lines) equal to the rest of the possibilities (or I forgot the deprecation warning in [Parser lines](https://invoice2data.readthedocs.io/latest/tutorial.html#parser-lines)).

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -230,7 +230,7 @@ If the first regexp for `VAT 10%` catches **1.5** and the second regexp for
 As you can see, the sum_ prefix is removed from the amount_tax field name.
 
 
-### Lines
+#### Lines
 
 The `lines` key allows you to extract line item data from invoices, such as product descriptions, quantities, and prices.
 


bosd:
Good catch — the deprecation being hidden in `Parser lines` only is a real gap. Two nits before this can merge:

1. If you demote `Lines` to `####`, its own subsections `Handling Multiple Lines per Item` (line 284) and `Recipes & troubleshooting` (line 303) also need to drop from `####` → `#####` so the hierarchy stays consistent.
2. Even with the demotion, a reader deep-linking to `#lines` won't see the parent context. Consider adding an explicit warning at the top of the section, e.g.:
   ```markdown
   #### Lines

   > **Deprecated.** Use [`parser: lines`](#parser-lines) instead. This top-level `lines:` key is kept for backwards compatibility and will be removed in a future release.
   ```
   You could keep or drop the structural demotion — the callout is the load-bearing signal.

Also: please rebase onto master (currently `BEHIND`). Thanks for spotting this!

PhilippFeO:
> If you demote Lines to ####, its own subsections Handling Multiple Lines per Item (line 284) and Recipes & troubleshooting (line 303) also need to drop from #### → ##### so the hierarchy stays consistent.

Ups, should have done this in the first place. I glanced over it to fast. Thank you!

> Even with the demotion, a reader deep-linking to #lines won't see the parent context.

True. I'll go with the explicit warning. 🤝

PR #749, opened 2026-08-21
feat: add record values to lines and camelot

pjkersten:
## Summary

  Adds record-level `static` and `defaults` values to the modern `lines`
  parser and Camelot record output.

  - Keeps `static_<field>` and `<field>_default` as `lines` compatibility aliases.
  - Adds `no_newline_fields` and opt-in `append_on_line` to `lines`.
  - Does not add record semantics to scalar parsers or the legacy text tables plugin.

  ## Tests

  38 targeted tests passed, including lines, Camelot, tax lines and UNECE UoM.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -300,6 +300,35 @@ If a line matches `first_line`, it starts a new item. If it matches `last_line`,
 When using `first_line` and `last_line`, make sure that `first_line` is the most specific regex and `line` is the least specific.
 ````
 
+#### Static values and defaults per record
+
+The `lines` parser produces a list of records. Use `static` for values that
+belong on every extracted record, and `defaults` to fill a missing or empty
+capture. This avoids adding artificial capture groups for values that are known
+from the template:
+
+```yaml
+fields:
+  lines:
+    parser: lines
+    start: 'Item'
+    end: 'Total'
+    line: '(?P<name>.+)\s+(?P<price_subtotal>\d+[.,]\d{2})'
+    static:
+      taxes:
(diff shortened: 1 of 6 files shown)


bosd:
Thanks @pjkersten — clean approach, especially the `records.py` centralization so lines and camelot share the same static/defaults semantics. I pushed a small fix commit (`9a10311`) covering two things I ran into while reviewing; feel free to squash or amend:

1. **`parse_current_row` docstring** — the pre-commit ruff hook was failing on D417 because the new `no_newline_fields` arg wasn't listed under Args. Added it. (Also flipped the docstring to `r"""` because ruff D301 flagged the `\n` reference.)
2. **`static: null` / `defaults: null` guard.** `settings.get("static", {})` returns `None` when someone writes bare `static:` in a YAML template, which then blows up on `.update(None)`. Swapped both to `settings.get("static") or {}`. Regression test added.

Two smaller things I noticed while reading, only worth touching if you feel like it:

- The `append_on_line: true` × `last_line: ...` interaction — when both are set, `last_line` still branches first, so `append_on_line` only fires for interior `line` matches. Might be worth a sentence in the tutorial so a template author doesn't get surprised.
- Tutorial example uses `taxes: [{amount: 21.0}]` — the canonical Odoo-oriented per-line schema uses `line_tax_percent`, so a downstream Odoo importer wouldn't map that shape. Fine for demonstrating the mechanism; consider `line_tax_percent: 21` for realism.

Otherwise nice PR — pattern generalizes cleanly to any future record-producing extractor.

PR #751, opened 2026-08-21
Add pennyocr input backend (hosted OCR API, stdlib-only)

imiraoui:
Adds a new input backend for [PennyOCR](https://pennyocr.com), a hosted VLM-based OCR API.

**Why it fits invoice2data:** users without a local OCR stack (no tesseract binary, no GPU for docTR/paddle, no GCS bucket for gvision) get scanned-invoice support with nothing but an env var. Handles scanned PDFs, photos and images; multipage output uses form-feed separators like the pdftotext backend, so existing templates behave identically.

**Implementation notes**
- Standard library only (urllib multipart) — zero new dependencies
- Follows `input/__interface__.py`: `to_text(path, area_details=None, **kwargs)`, `SUPPORTS_AREA = False`, `is_available()` gated on `PENNYOCR_API_KEY` so the backend self-excludes when unconfigured
- Registered in `INPUT_MODULES` as `pennyocr`; README list updated
- Unit tests included (`tests/test_pennyocr.py`, network mocked): registration, availability gating, multipart request shape, error path — all passing

**Service details, for transparency:** commercial API, $0.75 per 1,000 pages with 100 free pages/month (no card), zero data retention by default. Disclosure: I'm the maintainer of PennyOCR. Happy to adjust anything — naming, error semantics, or moving it behind an extra if you prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ volume of invoices, saving time and reducing manual-entry errors. It:
 1. extracts text from PDF files with a pluggable, cascading backend —
    `pdfium` (default, no system deps), `pdftotext`, `text`, `pdfminer`,
    `pdfplumber`, or OCR (`tesseract`, `ocrmypdf`, `docTR`, `paddleocr`,
-   `gvision`).
+   `gvision`, `pennyocr`).
 2. searches for regex in the result using a YAML or JSON-based template system
    (with an optional [AI fallback](https://invoice2data.readthedocs.io/en/latest/ai.html)).
 3. saves results as CSV, JSON or XML, or renames PDF files to match the content.
--- a/src/invoice2data/input/__init__.py
+++ b/src/invoice2data/input/__init__.py
@@ -19,6 +19,7 @@ from . import pdfminer_wrapper
 from . import pdfoxide
 from . import pdfplumber
 from . import pdftotext
+from . import pennyocr
 from . import tesseract
 from . import text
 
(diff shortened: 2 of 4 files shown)


bosd:
Thanks for the disclosure and the clean stdlib-only implementation. A couple of things before we can consider merging:

**Please move it behind an opt-in extra.** Bundled input backends in the default install list carry an implicit endorsement, and PennyOCR is a commercial hosted service. The right shape for this is:

- `[project.optional-dependencies]` extra called `pennyocr` (even if it declares zero deps — it still gates the README/docs surface).
- `is_available()` stays as-is (env-var gated), so the backend also self-excludes at runtime without the key.
- README + `docs/api.md` mention it as an *opt-in* backend, in the same tier as `gvision` / `documentai` / `doctr` / `paddleocr`, not in the always-on list.
- No entry in `INPUT_MODULES` unless the extra is available, or (simpler) leave it in `INPUT_MODULES` and just skip its README appearance in the default list.

That way users choosing it are opting in explicitly, and the project doesn't ship a paid-service pointer to everyone by default.

**CI is red on `master`:**
- `pre-commit` (ruff): two errors in `src/invoice2data/input/pennyocr.py` — please run `pre-commit run --all-files` locally.
- `mypy` (3.10 + 3.13): missing type annotations on `pennyocr.py:38`, `pennyocr.py:89`, and throughout `tests/test_pennyocr.py`. See the failing job for the full list.
- Branch is behind master; please rebase after fixing.

Once the extra + CI are addressed, we can take another look. Thanks!

PR #752, opened 2026-08-21
fix(loader): reject non-string keyword entries with a helpful hint (#743)

### LINKED ISSUE #743: Wrong template matching if one item in `keywords` or `exclude_keywords` ends on `:`
Maybe related to #742 (depending on the fix).

When I use an `exclude_keywords` ending on `:` (fi. `invoice id:`), I get the following error:

`TypeError: 'in <string>' requires string as left operand, not dict` at

Image

I am pretty sure, this originates from the `super` call to `OrderedDictType` in `__init__` parsing such an item as `dict` and not as `str`.

---

Some ideas on fixing this:
1. Stating in the docs that this will fail.
2. Removing trailing `:`.
3. (Unsure, since this seems to be related to `OrderedDictType`) Enabling regex in `keywords` and `exclude_keywords`.
4. Inserting manually, ie. circumvent `OrderedDictType` on these.

bosd:
Closes #743.

## Summary

When a template author writes an unquoted YAML keyword ending in `:`:

\`\`\`yaml
exclude_keywords: [invoice id:]
\`\`\`

`yaml.safe_load` parses that element as an inline **mapping** (`[{'invoice id': None}]`), not a string. The template then loaded silently, and later crashed deep inside `matches_input` with:

\`\`\`
TypeError: 'in <string>' requires string as left operand, not dict
\`\`\`

because the code runs `keyword in extracted_str`. The stack pointed at `OrderedDictType.__init__`, sending @PhilippFeO on a wild goose chase far from the YAML-quoting root cause.

## Fix

Catch it at `prepare_template` time: iterate `keywords` / `exclude_keywords`, reject any non-string entry with a warning that names the template, the field, the observed type and the fix (`- 'invoice id:'`). Returning `None` makes `read_templates` / `ordered_load` drop the offending template — same shape as the existing "missing keywords" bail-out just above.

## Behavior

Before:
\`\`\`
[silently loads] → TypeError deep in matches_input
\`\`\`

After:
\`\`\`
WARNING: Failed to load template trap.yml. `exclude_keywords` entry must be
a string, got dict ({'invoice id': None}). Hint: a YAML keyword ending in
`:` needs to be quoted, e.g. `- 'invoice id:'`.
\`\`\`

Non-breaking: any template that already loaded (all 215 bundled templates + every real user template that avoided this quoting trap) is unaffected.

## Test plan

- [x] `pytest tests/test_loader.py` — 18/18 (16 existing + 2 new)
  - `test_unquoted_yaml_keyword_ending_in_colon_is_rejected` — reproduces the #743 case and asserts the warning
  - `test_quoted_yaml_keyword_ending_in_colon_loads_normally` — locks that the properly-quoted form works
- [x] `pytest tests/test_template_rot.py` — 861 assertions across all 215 bundled templates, still green
- [x] `ruff check` / `ruff format --check` — clean

### DIFF
--- a/src/invoice2data/extract/loader.py
+++ b/src/invoice2data/extract/loader.py
@@ -249,6 +249,26 @@ def prepare_template(tpl: dict[str, Any]) -> dict[str, Any] | None:
     else:
         tpl["exclude_keywords"] = []  # Set to empty list if not present
 
+    # Issue #743: catch the YAML-quoting trap where a keyword ending in ``:``
+    # is parsed as an inline mapping ({'invoice id': None}) instead of a
+    # string. Left unchecked, ``matches_input`` later crashes with
+    # ``TypeError: 'in <string>' requires string as left operand, not dict``.
+    # Drop the offending template and log a hint that the value must be
+    # quoted in YAML.
+    for key in ("keywords", "exclude_keywords"):
+        for value in tpl[key]:
+            if not isinstance(value, str):
+                logger.warning(
+                    "Failed to load template %s. `%s` entry must be a string, "
+                    "got %s (%r). Hint: a YAML keyword ending in `:` needs to "
+                    "be quoted, e.g. `- 'invoice id:'`.",
+                    tpl.get("template_name", "<stream>"),
+                    key,
+                    type(value).__name__,
+                    value,
+                )
+                return None
+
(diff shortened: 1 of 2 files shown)


PhilippFeO:
❤️

PR #753, opened 2026-08-21
fix(#742): match keywords as regex, matching the documented tutorial

### LINKED ISSUE #742: Mismatch between docs and code
In [Tutorial for Template Creation > Simple Invoice Template](https://invoice2data.readthedocs.io/latest/tutorial.html#simple-invoice-template) `exclude_keyword` is defined as: "Optional. Regex patterns that, if matched, prevent the template from being used." and the example also uses a Regex pattern. Similar for `keywords`: "These keywords are regex patterns, so ‘Company US’ and ‘Company\s+US’ both work. This also allows some flexibility as ‘Company\s+(US|UK)’ […]".

But in the code, no regex logic is applied, only a naive `in` test with strings:

Image

The docstring is (depending on interpretation) more explicit here, because it doesn't mention "regex" at all and only refers to strings.

---

In my opinion there are two (obvious) fixes:

1. Updating documentation.
2. Use regex matching in the snippet above.

bosd:
## Summary

Closes #742.

The template-authoring tutorial has always documented `keywords` and `exclude_keywords` as **regex patterns**  e.g. \"`Company US` and `Company\s+US` both work\". But `InvoiceTemplate.matches_input` used a plain `in` substring test, so any keyword that leaned on a regex metachar silently under-matched.

An audit of the bundled templates shows ~16 already write regex-shaped keywords that were effectively broken:

- `(?i)Accor`, `(?i)Digital\sRiver`
- `BE\s0673\s923`
- `Express\s+\w+`
- `Wystawił\(a\)`
- `NL8152[.]54[.]295[.]B01`
- …and more.

## Change

- Route both `keywords` and `exclude_keywords` checks through a new module-local `_keyword_matches` helper that calls `_regex.search` (the compile-once cache used everywhere else in `extract/`).
- If a keyword doesn't compile as valid regex, fall back to a literal substring check and DEBUG-log the fallback  so a stray unescaped `[` or `(` in a simple-string keyword like `Company [Ltd` keeps matching literally instead of breaking.
- Re-export `_regex.error` (the active engine's error type  `re.error` by default, `regex.error` under `INVOICE2DATA_REGEX_ENGINE=regex`) so the fallback doesn't need to know which engine is active.

## Compatibility

- Plain-string keywords with no metacharacters (`Acme Corp`) behave exactly as before — `regex.search(\"Acme Corp\", text)` is the same as `\"Acme Corp\" in text`.
- Keywords that already were regex (bundled templates) now actually work.
- Keywords that were **string-but-look-like-regex** (rare  `Company [Ltd` with an unclosed `[`) still match via the literal fallback.

## Test plan

- [x] New unit tests for regex `\s+`, `(?i)`, alternation, exclude-as-regex, and invalid-regex-falls-back-to-substring.
- [x] Existing test suite passes (only the pre-existing env-related failures remain — `test_gvision`, `test_copy` needing pdftotext/tesseract, and `QualityHosting.pdf` cascade, all red on master too).
- [x] `test_default_templates_are_loaded` still loads every bundled template (nothing rejected as invalid regex).

### DIFF
--- a/src/invoice2data/extract/_regex.py
+++ b/src/invoice2data/extract/_regex.py
@@ -26,6 +26,11 @@ if os.environ.get("INVOICE2DATA_REGEX_ENGINE", "re").lower() == "regex":
 #: Name of the active regex engine ("re" or "regex").
 ENGINE: str = _engine.__name__
 
+#: The active engine's ``PatternError``/``error`` type, re-exported so callers
+#: can ``except _regex.error`` without caring which engine is active. Both
+#: engines expose the same attribute name.
+error = _engine.error
+
 
 @lru_cache(maxsize=4096)
 def compile(pattern: str, flags: int = 0) -> "re.Pattern[str]":
--- a/src/invoice2data/extract/invoice_template.py
+++ b/src/invoice2data/extract/invoice_template.py
@@ -32,10 +32,10 @@ optimized_str_logger = getLogger("invoice2data.optimized_str")
 
 
 def _keyword_matches(keyword: str, text: str) -> bool:
-    """Return whether ``keyword`` matches ``text`` as a regex (issue #742).
+    r"""Return whether ``keyword`` matches ``text`` as a regex (issue #742).
 
     The tutorial has always documented ``keywords`` / ``exclude_keywords`` as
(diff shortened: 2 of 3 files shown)


PhilippFeO:
🚀 ❤️

